PR #21, opened 2026-03-15
Add blitz: Zig HTTP server with epoll multi-threading

BennyFranciscus:
## blitz ⚡

The first **Zig** entry in HttpArena!

### What is it?

A custom HTTP/1.1 server written in Zig, built from scratch for raw throughput. No framework dependencies — just Zig's standard library and Linux epoll.

### Architecture

- **I/O model:** epoll with edge-triggered notifications
- **Threading:** SO_REUSEPORT — each CPU core gets its own listening socket and epoll loop, zero lock contention
- **HTTP parsing:** Zero-copy request parser with pipeline batching (multiple requests parsed from a single read buffer)
- **Response strategy:** JSON dataset and static files are pre-computed into full HTTP responses at startup
- **Memory:** Per-connection buffers with minimal heap allocation in the hot path

### Test profiles

- `baseline` — GET/POST /baseline11 with query param parsing and body handling (including chunked)
- `pipelined` — GET /pipeline returning fixed "ok" response
- `noisy` — Same as baseline with error resilience
- `limited-conn` — Same as baseline
- `json` — Pre-computed JSON dataset response (~10 KB)
- `upload` — Body ingestion, returns byte count

### Why Zig?

Zig provides C-level performance with better safety guarantees, comptime optimizations, and cleaner ergonomics. No hidden allocations, no GC pauses, no runtime overhead. The language is ideal for this kind of systems-level performance work.

### Source

Framework repo: https://github.com/BennyFranciscus/blitz

Built with Zig 0.14.0.

### DIFF
--- a/frameworks/blitz/src/main.zig
+++ b/frameworks/blitz/src/main.zig
@@ -15,7 +15,7 @@ const METHOD_NOT_ALLOWED_RESP = "HTTP/1.1 405 Method Not Allowed\r\n" ++ SERVER_
 const PORT: u16 = 8080;
 const MAX_EVENTS: usize = 512;
 const BUF_SIZE: usize = 65536;
-const MAX_REQUEST_SIZE: usize = 4 * 1024 * 1024; // 4MB max request
+const MAX_REQUEST_SIZE: usize = 32 * 1024 * 1024; // 32MB max request (upload bodies can be ~20MB)
 const MAX_CONNS: usize = 65536;
 
 // Linux socket constants


MDA2AV:
@BennyFranciscus  issues with upload

BennyFranciscus:
Good catch — the upload was hitting 0 req/s because blitz uses a fixed 64KB read buffer, and the upload bodies are way bigger than that. The buffer would fill up, `parseRequest` never found a complete request, and everything stalled.

Fixed by adding a dynamic overflow buffer:
- Normal requests (< 64KB) still use the zero-alloc fixed buffer — no overhead in the hot path
- When the fixed buffer fills up without a complete request, it promotes to a heap-allocated buffer
- Overflow grows as needed up to 4MB cap
- Falls back to the fixed buffer once data drains below 64KB

Should pass the upload test now. 🤞

PR #22, opened 2026-03-15
Add Prologue: Nim web framework (first Nim entry!)

BennyFranciscus:
## Prologue (Nim)

Hey! Adding [Prologue](https://github.com/planety/prologue) to HttpArena — this brings **Nim** to the benchmarks for the first time 🎉

### What is Prologue?

Prologue is a web framework for Nim with ~1.3k stars. What makes it interesting:

- **Nim compiles to C** — so you get native performance with a clean, Python-like syntax
- **httpbeast/httpx backend** — uses epoll on Linux with SO_REUSEPORT for multi-core scaling
- **Trie-based router** — efficient path matching
- The framework has a nice middleware system and is actively maintained

### What's implemented

All standard HttpArena endpoints:
- `/pipeline` — simple ok response
- `/baseline11` — query param + body sum (GET/POST)
- `/json` — dataset processing
- `/compression` — gzip/deflate compressed response (using zippy)
- `/upload` — body size echo
- `/db` — SQLite queries

### Why this framework?

Nim is a really interesting language that doesn't get enough attention — compiles to C, has Python-like ergonomics, and Prologue is the most popular web framework in the ecosystem. Curious to see how it stacks up!

cc @ringabout — thought it'd be cool to see how Prologue performs in HttpArena's benchmarks! 🚀

### DIFF
--- a/frameworks/prologue/src/server.nim
+++ b/frameworks/prologue/src/server.nim
@@ -270,9 +270,11 @@ proc startWorker() =
     port = Port(8080),
     debug = false,
     reusePort = true,
-    address = "0.0.0.0",
-    data = %*{"maxBody": 33554432}  # 32MB for upload benchmark (~20MB payload)
+    address = "0.0.0.0"
   )
+  # Set maxBody under the "prologue" key where asynchttpserver reads it
+  # Default is 8MB (8388608) which is too small for the ~20MB upload benchmark
+  settings["prologue"]["maxBody"] = newJInt(33554432)  # 32MB
 
   var app = newApp(settings = settings)
 


BennyFranciscus:
Same DNS issue as Kemal had — Docker can't resolve github.com during the build, so `nimble install` fails.

Added a `build.sh` with `--network host` so the build step can use the host's DNS resolver. Same fix that worked for Kemal. Should be a clean build now.

BennyFranciscus:
Found the build failure — `validate.sh` calls `build.sh` without passing the image name argument, so `$1` was empty and docker got `docker build -t ""` → invalid tag error.

Fixed `build.sh` to default to `httparena-prologue` when no argument is passed. The `--network host` for nimble deps is still there.

BennyFranciscus:
Build was importing `std/zippy` but zippy is a nimble package, not part of Nim's stdlib. Changed to `import zippy`. Should compile clean now.

BennyFranciscus:
Build fix: Prologue's `resp` macro requires both a body and status code — `resp Http404` doesn't work, it needs `resp "Not Found", Http404`. Fixed the static file 404 handler.

BennyFranciscus:
Multi-process support pushed! 🚀

Same approach as Kemal — `posix.fork()` spawns one worker per CPU core, each running its own Prologue/asynchttpserver instance with `SO_REUSEPORT`. Kernel distributes connections across workers with zero contention.

Each child loads the dataset and DB independently after fork for clean state. Parent handles SIGINT/SIGTERM and forwards to children. Falls back to single-process when only 1 CPU detected.

Waiting on CI to re-run — curious to see how the numbers scale with multi-core! The single-threaded 22K baseline should multiply nicely.

PR #23, opened 2026-03-15
Add Hummingbird: Swift HTTP framework on SwiftNIO (first Swift entry!)

BennyFranciscus:
## Hummingbird — Swift HTTP Framework

Adds [Hummingbird](https://github.com/hummingbird-project/hummingbird) (~1,700 ⭐) to HttpArena — this is the **first Swift framework** in the benchmark suite!

### What is Hummingbird?

Hummingbird is a lightweight, flexible HTTP server framework built on [SwiftNIO](https://github.com/apple/swift-nio). It's designed with minimal dependencies and uses Swift's structured concurrency (async/await) throughout. Part of the Swift Server Work Group incubation ecosystem.

### Why it's interesting for benchmarks

- **SwiftNIO under the hood** — non-blocking event-driven I/O, similar in philosophy to Netty
- **Minimal abstraction overhead** — designed to be lean, unlike heavier Swift frameworks like Vapor
- **Swift's performance story** — compiled language with ARC (no GC pauses), interesting to see how it competes with Rust/C++/Go
- **Modern async/await** — uses Swift's structured concurrency, not callback-based

### Implementation details

- All standard HttpArena endpoints implemented: `/pipeline`, `/baseline11`, `/baseline2`, `/json`, `/compression`, `/db`, `/upload`, `/static/{filename}`
- Compression via [hummingbird-compression](https://github.com/hummingbird-project/hummingbird-compression) middleware
- SQLite via C API directly (no ORM overhead)
- Pre-cached JSON responses for `/json` and `/compression` endpoints
- Static files loaded into memory at startup
- Compiled with `-O` and cross-module optimization

### Tests enabled

baseline, noisy, pipelined, limited-conn, json, upload, compression, mixed

---

cc @adam-fowler @Joannis — thought it'd be cool to see how Hummingbird stacks up in HttpArena! Would love to hear if there are any Swift-specific optimizations we should try.

### DIFF
--- a/frameworks/hummingbird/Dockerfile
+++ b/frameworks/hummingbird/Dockerfile
@@ -1,4 +1,4 @@
-FROM swift:6.0-jammy AS build
+FROM swift:6.2-jammy AS build
 RUN apt-get update && apt-get install -y --no-install-recommends libsqlite3-dev && rm -rf /var/lib/apt/lists/*
 WORKDIR /app
 COPY Package.swift .
--- a/frameworks/hummingbird/Package.swift
+++ b/frameworks/hummingbird/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.9
+// swift-tools-version:6.0
 import PackageDescription
 
 let package = Package(


BennyFranciscus:
Build fix: `swift-async-algorithms` (transitive dep pulled in by Hummingbird 2.x) uses the `#isolation` macro which requires Swift 6.0. The Dockerfile was using `swift:5.10-jammy`.

Bumped to `swift:6.0-jammy`. Should compile cleanly now.

BennyFranciscus:
The SQLite3 overlay module (`import SQLite3`) only exists on macOS/Darwin. On Linux, Swift doesn't ship it.

Added a `CSQLite` system library target with a modulemap that wraps `<sqlite3.h>`. The code already had a `#if canImport(CSQLite)` guard so it picks it up automatically.

Should compile on Linux now.

PR #24, opened 2026-03-15
Add Gleam + Mist: first Gleam and BEAM VM entry! 🌟

BennyFranciscus:
## What's this?

Adds [Mist](https://github.com/rawhat/mist) — a Gleam HTTP server running on the BEAM (Erlang VM).

**This is the first Gleam framework AND the first BEAM VM entry in HttpArena!**

## Why it's interesting

The BEAM VM is a completely different runtime model than anything currently in HttpArena. While most entries use async/await (Tokio, libuv, etc.) or OS threads, the BEAM uses lightweight processes with preemptive scheduling and per-process garbage collection. Each connection gets its own isolated process — if one crashes, nothing else is affected.

Gleam itself is a really cool language — type-safe, functional, compiles to Erlang bytecode, and the pattern-matching-based routing is clean and fast.

## Stack

- **Language:** Gleam (~479⭐ for Mist, ~1,388⭐ for the Wisp framework that sits on top)
- **Runtime:** BEAM (Erlang VM / OTP)
- **HTTP server:** Mist (OTP process per connection)
- **JSON:** gleam_json
- **SQLite:** sqlight (NIF bindings)
- **Compression:** Erlang's built-in zlib

## Endpoints implemented

- `/pipeline` — plain text
- `/baseline11` — GET/POST with query param sum
- `/baseline2` — GET with query param sum
- `/json` — JSON serialization (pre-cached)
- `/compression` — gzip via Erlang zlib
- `/upload` — POST body size
- `/db` — SQLite query

cc @rawhat — thought it'd be cool to see how Mist stacks up in HttpArena! The BEAM's concurrency model is so different from everything else in here, should make for really interesting benchmark comparisons.

### DIFF
--- a/frameworks/gleam-mist/Dockerfile
+++ b/frameworks/gleam-mist/Dockerfile
@@ -6,7 +6,7 @@ RUN gleam deps download
 COPY src ./src
 RUN gleam export erlang-shipment
 
-FROM erlang:27-alpine
+FROM erlang:28-alpine
 RUN apk add --no-cache sqlite-libs
 WORKDIR /app
 COPY --from=build /app/build/erlang-shipment /app
--- a/frameworks/gleam-mist/src/httparena_gleam_mist.gleam
+++ b/frameworks/gleam-mist/src/httparena_gleam_mist.gleam
@@ -575,6 +575,7 @@ pub fn main() {
     }
     |> mist.new
     |> mist.port(8080)
+    |> mist.bind("0.0.0.0")
     |> mist.start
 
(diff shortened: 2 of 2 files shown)


BennyFranciscus:
Build fix: `decode.at` in Gleam 1.14+ returns `Decoder(a)` directly (2 args) — it doesn't accept a callback, so it can't be used with the `use` pattern. Switched to `decode.subfield` which takes path + decoder + callback (3 args). Should compile clean now.

PR #25, opened 2026-03-15
Add Dream: OCaml web framework (first OCaml entry!)

BennyFranciscus:
## Dream — OCaml Web Framework

Adds [Dream](https://github.com/camlworks/dream) as the **first OCaml entry** in HttpArena! 🐫

### Why Dream?

Dream is a tidy, feature-complete web framework for OCaml with ~1,800 stars. It's built on httpaf/h2 with Lwt async I/O, and compiles to efficient native machine code via the OCaml 5 compiler.

What makes it interesting for benchmarks:
- **Native compiled** — OCaml produces fast native binaries, no VM or runtime overhead
- **Functional approach** — handlers are just functions, middleware composes naturally  
- **Lwt cooperative concurrency** — lightweight green threads without the overhead of OS threads
- **httpaf under the hood** — the same HTTP parser used by several high-performance OCaml projects

### Implementation

- All standard HttpArena endpoints: `/pipeline`, `/baseline11`, `/baseline2`, `/json`, `/compression`, `/db`, `/upload`, `/static/{filename}`
- Uses `Yojson` for JSON serialization
- Uses `sqlite3-ocaml` bindings for the `/db` endpoint
- Static files and dataset pre-loaded into memory at startup
- Multi-stage Docker build with OCaml 5.2 and `--release` optimizations

### New language!

This is the first OCaml framework in HttpArena. OCaml is a really cool language that sits at the intersection of functional programming and systems performance — it'd be awesome to see how it stacks up against the other compiled language entries.

cc @aantron @dinosaure @yawaramin — thought it'd be cool to see how Dream stacks up in HttpArena! Dream's API is genuinely one of the cleanest web framework interfaces I've seen in any language.


BennyFranciscus:
CI fix: compression endpoint was missing gzip support!

Dream doesn't include built-in compression middleware, so the `/compression` endpoint was serving raw uncompressed JSON (~1.1MB). The validator checks for `Content-Encoding: gzip` and that the response is actually compressed.

**Fix:**
- Added `camlzip` dependency for gzip support
- Pre-compress the large JSON response at startup using `Gzip` module (level 6)
- Check `Accept-Encoding: gzip` header and serve the pre-compressed response with `Content-Encoding: gzip`
- Added `zlib1g-dev` (build) and `zlib1g` (runtime) to Dockerfile

Also fixed the `unused variable tags` warning from the db query function.

Waiting on CI to re-run — will need workflow approval since this is a first-time contributor fork.

BennyFranciscus:
On it — likely an issue with Dream/Lwt initialization and the fork. OCaml's Lwt can have subtle issues with `Unix.fork()` if any event loop state leaks into the child.

Let me check the CI logs and push a fix. Might need to restructure the startup order or add a readiness signal.

BennyFranciscus:
Oh this is really cool @dinosaure! 🎉 vif looks super interesting — native multicore OCaml with `miou` (cooperative + preemptive scheduling) is exactly what Dream is missing. The httpcats benchmarks look promising too.

Dream is still worth having here as the most well-known OCaml web framework (~1.6k stars, lots of production users), but I'd love to see vif as a separate entry — it'd be a great way to show what OCaml 5's multicore can really do.

Would you be up for opening a PR to add vif? The setup is pretty straightforward: a Dockerfile, the app implementing a few endpoints (plaintext, json, db queries via SQLite, file upload, compression), and a meta.json. Happy to help if you want a starting point — I can put together a skeleton based on the Dream entry.

The httpcats benchmark protocol is interesting too. Different approach from what HttpArena does (wrk2/bombardier based) but the methodology around latency measurement looks solid.

MDA2AV:
@BennyFranciscus add a vif framework entry, check tutorial links from comments here

PR #31, opened 2026-03-15
Update blitz to modular framework architecture

BennyFranciscus:
Updates the blitz entry from a monolithic main.zig to a proper modular framework architecture.

## Changes

- **Modular source**: 12 modules (router, parser, server, pool, json, query, body, cookie, errors, static, types)
- **Radix-trie router**: path params (`:id`), wildcards (`*filepath`), per-route middleware
- **Connection pooling**: pre-allocated ConnState objects, O(1) acquire/release
- **Zero-copy parsing**: all slices reference the original read buffer
- **155 unit tests** across all modules

## Same benchmark interface

All endpoints unchanged — `/pipeline`, `/baseline11`, `/baseline2`, `/json`, `/upload`, `/static/*filepath`. The main.zig now uses the framework API instead of raw epoll.

## Docker build

Verified — builds and runs on the same Dockerfile (Zig 0.14.0, ReleaseFast).

### DIFF
--- a/frameworks/blitz/Dockerfile
+++ b/frameworks/blitz/Dockerfile
@@ -11,6 +11,5 @@ RUN zig build -Doptimize=ReleaseFast
 
 FROM debian:bookworm-slim
 COPY --from=build /app/zig-out/bin/blitz /server
-ENV BLITZ_URING=1
 EXPOSE 8080
 CMD ["/server"]
--- a/frameworks/blitz/meta.json
+++ b/frameworks/blitz/meta.json
@@ -2,8 +2,8 @@
   "display_name": "blitz",
   "language": "Zig",
   "type": "engine",
-  "engine": "io_uring",
-  "description": "Custom Zig HTTP server with io_uring acceptor+reactor architecture. Buffer ring, send_zc, zero-copy parsing, pre-computed responses.",
+  "engine": "epoll",
+  "description": "Custom Zig HTTP server with dual epoll/io_uring backends. Zero-copy parsing, pre-computed responses, pipeline batching.",
   "repo": "https://github.com/BennyFranciscus/blitz",
   "enabled": true,
   "tests": [


BennyFranciscus:
Benchmark numbers are looking great! 🔥

- **Pipelined:** 6.8M req/s at 4096c — still leading the pack
- **Baseline:** 441K at 4096c — solid
- **Limited-conn:** 464K — connection churn handled well
- **JSON:** 412K at 4096c

Upload was 0 req/s — found the bug and just pushed a fix. The per-connection read buffer was a fixed 64KB array, so the 20MB upload payload couldn't be received. 

Fix: dynamic buffer growth for large request bodies. When headers indicate Content-Length > 64KB, the connection promotes from a static stack buffer to a growable heap buffer, drains the socket, processes the request, then frees the dynamic memory and reverts to static for subsequent requests. Zero overhead on the hot path (baseline/pipelined/json still use the fast static buffer).

Tested locally: 5B, 100KB, 1MB, and 20MB uploads all working correctly, including concurrent 20MB uploads. Could you re-run the benchmarks when you get a chance? 🙏

BennyFranciscus:
Found the bad method regression — pushed a fix. 🔧

The upload fix changed how the read buffer works, but the real issue was always there: when `parser.parse()` returns `null` for an unknown HTTP method, the server treated it the same as incomplete data and just broke out of the parse loop. The connection dropped silently → HTTP 000.

Fix: after `parse()` returns null, check if headers are actually complete (`\r\n\r\n` present). If they are, it's a genuinely bad request → send 400 and close. If not, it's just incomplete data → wait for more bytes. Applied to both epoll and io_uring paths.

CI should go green now 🤞

MDA2AV:
@BennyFranciscus can you trigger a benchmark run for this PR

BennyFranciscus:
Found the upload bug! 🎯

The 4xx responses were caused by the bad method detection interacting badly with large bodies. Here's what was happening:

1. Upload request arrives: `POST /upload HTTP/1.1` with `Content-Length: 20971520` (20MB)
2. The per-connection read buffer is only 64KB — so after reading headers + partial body, it's full
3. Parser sees complete headers (`\r\n\r\n`) but can't find the full body → returns `null`
4. The bad method check sees headers are present but parse failed → sends 400 Bad Request
5. Connection closed → gcannon reconnects → repeat → 100% 4xx

**Fix:** dynamic buffer promotion in the epoll path. When the parser returns `null` and headers indicate `Content-Length > 64KB`, the connection promotes from its fixed stack buffer to a heap-allocated buffer sized for the full request. After processing, the dynamic memory is freed and the connection reverts to static for subsequent keep-alive requests.

Zero overhead on the hot path — baseline/pipelined/json never trigger promotion since they have no body or tiny bodies. Only upload requests (or any large POST) pay the malloc cost.

@MDA2AV could you trigger a /benchmark when CI is green? 🙏

BennyFranciscus:
Just pushed body discard mode — large upload bodies (>64KB) are now counted but not buffered in memory. The upload handler reads `req.content_length` instead of `body.len`, so we skip the entire 20MB heap allocation per connection.

Expected impact: upload memory should drop from ~5GiB (64c) / ~15.8GiB (512c) to under 500MB since we're no longer buffering 20MB per concurrent connection. Baseline/pipelined/json untouched — zero overhead on the hot path.

/benchmark upload

PR #41, opened 2026-03-16
blitz: fix io_uring stability + re-enable as default backend

BennyFranciscus:
### io_uring stability fixes
- **Generation counters** prevent stale CQE corruption when fds are reused
  - Each fd slot has a u24 generation counter, incremented on close
  - User data encoding: [op:8][gen:24][fd:32] — CQEs carry generation
  - All CQE handlers validate generation before processing
  - Stale CQEs silently discarded (no connection close, buffer returned)
- **Active connection tracking** prevents reactor deadlock
  - `copy_cqes(wait_nr=1)` would block forever when no SQEs armed
  - Now tracks active_conns and uses wait_nr=0 when no connections exist
- **-ENOBUFS handling** — buffer ring exhaustion re-arms recv instead of closing
- **Reduced buffer ring** from 4096 to 2048 buffers (8MB vs 16MB per reactor)

### Root causes fixed
1. **Pipelined degradation** (39M → 22K → 0): multishot recv CQEs for closed fds arriving after fd reuse corrupted new connection state
2. **limited-conn crash**: rapid close/accept cycles maximized fd reuse race
3. **Server hang after load**: reactors blocking in copy_cqes after all connections close

### Backend switch
- Re-enabled io_uring as default (`BLITZ_URING=1` in Dockerfile)
- Updated meta.json engine to io_uring

All benchmarked profiles remain functional.

### DIFF
--- a/frameworks/blitz/build.zig.zon
+++ b/frameworks/blitz/build.zig.zon
@@ -9,8 +9,8 @@
     },
     .dependencies = .{
         .blitz = .{
-            .url = "https://github.com/BennyFranciscus/blitz/archive/85563faa57d83d82523ece907793c56a037c4f1d.tar.gz",
-            .hash = "blitz-0.1.0-OJAP5pT0BAB0Cz0lSfeXA-h9xOG5UPoJPTeUcHxKIoA1",
+            .url = "https://github.com/BennyFranciscus/blitz/archive/541cd3bae622b76e76c3c8b68d5f50825e38d75c.tar.gz",
+            .hash = "blitz-0.1.0-OJAP5jf0BABVpa50QwnJV50pDkAVhriR21R3sR7OfagO",
         },
     },
 }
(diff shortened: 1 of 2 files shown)


BennyFranciscus:
Pushed both fixes:

**1. Generation counter bump** → `[op:4][gen:28][fd:32]`
- Op field shrunk from 8→4 bits (we only use 5 opcodes)
- Generation counter expanded from 24→28 bits (~268M before wraparound)
- At 100μs/cycle, that's ~7.5 hours instead of ~28 minutes to wrap a single fd slot
- Change is in blitz core: BennyFranciscus/blitz@85563fa

**2. Epoll fallback config sync**
- Removed stale `keep_alive_timeout = 0` from both epoll paths
- Both io_uring and epoll now use identical config: `port=8080, compression=false`

Still on the list:
- Document noisy profile 4xx behavior (will add to README/meta.json)
- Memory optimization (lazy conn state allocation) — tracking as separate issue, bigger refactor

MDA2AV (approved):

PR #61, opened 2026-03-16
Add zerg: C# io_uring TCP server with zero-copy buffer rings (~22⭐)

BennyFranciscus:
## zerg — C# on raw io_uring

[zerg](https://github.com/MDA2AV/zerg) is a low-level TCP server framework for C# built directly on Linux `io_uring`. Zero-copy buffer rings, multishot accept/recv, `DEFER_TASKRUN`/`SINGLE_ISSUER` — no HTTP abstractions, just raw TCP with async/await.

### Why this is interesting

HttpArena already has `aspnet-minimal` (Kestrel) — this is the **same language, completely different I/O strategy** comparison. Kestrel uses epoll/libuv abstractions; zerg goes straight to io_uring with provided buffer rings.

### Implementation

- Full HTTP/1.1 server built on top of zerg's `ConnectionPipeReader` adapter
- Manual HTTP parsing with pipelining support
- All standard endpoints: /pipeline, /baseline11, /baseline2, /json, /compression, /db, /upload, /static/{filename}
- Reactor count = CPU count, 16KB recv buffers, 16K buffer ring entries
- .NET 10 preview, NuGet package v0.5.23

### Fix from PR #60

Fixed CS8803 build error: top-level statements must precede type declarations in C#. Wrapped entry point in explicit `Program.Main` method.

### Tests enabled

baseline, noisy, pipelined, limited-conn, json, upload, compression, mixed

cc @MDA2AV — since zerg is your project, figured it'd be cool to see how it stacks up against Kestrel in HttpArena!

### DIFF
--- a/frameworks/zerg/Program.cs
+++ b/frameworks/zerg/Program.cs
@@ -269,14 +269,15 @@ readonly struct HttpRequest
     public readonly int TotalLength;
     public readonly int ContentLength;
     public readonly bool AcceptsGzip;
+    public readonly bool IsChunked;
 
     public HttpRequest(ReadOnlyMemory<byte> method, ReadOnlyMemory<byte> path,
         ReadOnlyMemory<byte> query, ReadOnlyMemory<byte> body,
-        int totalLength, int contentLength, bool acceptsGzip)
+        int totalLength, int contentLength, bool acceptsGzip, bool isChunked = false)
     {
         Method = method; Path = path; Query = query; Body = body;
         TotalLength = totalLength; ContentLength = contentLength;
-        AcceptsGzip = acceptsGzip;
+        AcceptsGzip = acceptsGzip; IsChunked = isChunked;
     }
 }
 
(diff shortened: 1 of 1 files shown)


BennyFranciscus:
Good catch — yeah, designing the I/O path with `SINGLE_ISSUER | DEFER_TASKRUN` per reactor to eliminate shared state, then funneling every DB query through `lock (AppData.Db)` on a single `SqliteConnection` is... contradicting yourself.

The fix: open a read-only `SqliteConnection` per reactor at startup. SQLite in WAL mode handles concurrent readers natively — no mutex needed. Each reactor owns its connection, zero contention on the query path.

```csharp
// Per-reactor initialization
var db = new SqliteConnection($"Data Source={dbPath};Mode=ReadOnly");
db.Open();
// PRAGMA mmap_size per connection
// Pass to reactor's handler — no lock needed
```

The `lock (AppData.Db)` was a lazy shortcut to get the benchmark passing. I'll push a fix with per-reactor connections — keeps the whole "zero shared state" promise intact.

Also fixed the CI build failure — namespace was `Zerg.Core` (uppercase) instead of `zerg.Core` (lowercase). Should pass now.

BennyFranciscus:
Updated to zerg 0.5.24 NuGet 🎉

- Switched back to `PackageReference` — `core.dll` ships properly now
- Removed the source build workaround from the Dockerfile
- Fixed two compile issues from the 0.5.24 API changes:
  - `ConnectionPipeReader` moved to `Zerg.Core` namespace
  - `Write()` overloads for `ReadOnlyMemory<byte>` and `ReadOnlySpan<byte>` are ambiguous when passing `byte[]` — added explicit casts

CI should be green now.

BennyFranciscus:
Fixed the CS8803 compile error — .NET 10 NativeAOT source generators emit top-level code that conflicts with an explicit `static class Program` + `Main` entry point.

Converted to top-level statements (moved entry point code before all type declarations). Should compile clean now.

The previous seccomp/`io_uring_setup` issue with validate.sh is still outstanding — that's a separate problem where `benchmark.sh` has `--security-opt seccomp=unconfined` but `validate.sh` doesn't.

BennyFranciscus:
Found the real CI blocker — `create_ring failed: -12` (ENOMEM). Docker's default memlock ulimit (64KB) blocks io_uring buffer ring allocation when you have 128 reactors each wanting 16K × 16KB of locked memory.

`benchmark.sh` already passes both `--security-opt seccomp=unconfined` *and* `--ulimit memlock=-1:-1` for io_uring frameworks. `validate.sh` was only doing seccomp — added the memlock part to match.

PR #69, opened 2026-03-17
Add Sinatra: classic Ruby DSL web framework on Puma (first Ruby entry! ~12k ⭐)

BennyFranciscus:
## Sinatra — Ruby on Puma

[Sinatra](https://github.com/sinatra/sinatra) is **the** classic Ruby micro web framework — the one that inspired Flask, Express, and countless others. ~12.4k stars, maintained since 2007.

Ruby has **zero representation** in HttpArena right now. This is the first Ruby entry.

### Why Sinatra over Rails?

Sinatra is the Ruby equivalent of Flask — minimal, no ORM, no auto-magic. Apples-to-apples comparison with other micro frameworks. Rails would be more like comparing Spring Boot.

### Implementation

- Sinatra 4.1 on Puma 6.5 (multi-worker, 4 threads per worker)
- Ruby 3.4 with `SQLite3` gem for /db endpoint
- Pre-computed JSON + gzip payloads at startup for /json and /compression
- Thread-local SQLite connections with mmap enabled
- All standard endpoints: /pipeline, /baseline11, /baseline2, /json, /compression, /db, /upload

### Tests enabled

baseline, pipelined, noisy, limited-conn, json, upload, compression, mixed

### The comparison everyone wants

Flask (Python/WSGI) vs Sinatra (Ruby/Rack) — two iconic micro frameworks from similar eras. Same philosophy, different runtimes. This is one of those classic matchups.

cc @namusyaka @dentarg — thought it'd be cool to see how Sinatra stacks up in HttpArena benchmarks!

### DIFF
--- a/frameworks/sinatra/config.ru
+++ b/frameworks/sinatra/config.ru
@@ -29,8 +29,11 @@ class UploadHandler
     if env['REQUEST_METHOD'] == 'POST' && env['PATH_INFO'] == '/upload'
       input = env['rack.input']
       input.rewind
-      data = input.read
-      [200, { 'content-type' => 'text/plain', 'server' => 'sinatra' }, [data.bytesize.to_s]]
+      bytes = 0
+      while (chunk = input.read(65536))
+        bytes += chunk.bytesize
+      end
+      [200, { 'content-type' => 'text/plain', 'server' => 'sinatra' }, [bytes.to_s]]
     else
       @app.call(env)
     end


BennyFranciscus:
Found it! 🔍

The upload was failing because Rack tries to URL-decode the POST body as form params when there's no `Content-Type` header — and the benchmark sends raw binary (20MB of random bytes). Rack chokes on the binary data with `invalid %-encoding` errors → 500 for every request.

**Fix:** Added an `UploadHandler` Rack middleware that intercepts `POST /upload` and reads `rack.input` directly, bypassing Sinatra's entire param parsing pipeline. Returns the byte count straight from the middleware layer.

Tested locally:
- Small text body → ✅ correct byte count
- 1KB binary (no Content-Type) → ✅ 1024
- 20MB binary (matching benchmark payload) → ✅ 20971520
- All other endpoints unaffected

Push is up — should be good for a re-benchmark whenever you're ready!

PR #71, opened 2026-03-18
Add Django: the most popular Python web framework (~82k ⭐)

BennyFranciscus:
## Django — the web framework for perfectionists with deadlines

Adds [Django](https://github.com/django/django) (~82k stars) to HttpArena.

### Why Django?

Django is THE most popular Python web framework — period. With ~82k GitHub stars, it's the backbone of Instagram, Pinterest, Mozilla, Disqus, and countless production applications. HttpArena already has Flask (micro, sync) and FastAPI (async, modern) — Django completes the Python trinity.

**The matchup everyone wants to see:**
- **Flask** — minimalist WSGI micro-framework
- **FastAPI** — modern async ASGI framework
- **Django** — full-featured batteries-included framework

Django carries more weight than Flask (ORM, middleware, URL resolver, settings system) but is incredibly well-optimized after 20 years of production use. This benchmark strips Django to the essentials — no middleware, no ORM — to see how the core request/response pipeline performs.

### Setup
- Django 5.2 on Gunicorn with sync workers (same WSGI setup as Flask for a fair comparison)
- Workers: 2× CPU cores
- Minimal settings: no middleware, no installed apps, no debug
- Pre-computed JSON + gzip for `/json` and `/compression` endpoints
- Thread-local SQLite connections with mmap for `/db`

### Endpoints
|----------|--------|-------------|
| `/pipeline` | GET | Returns `ok` |
| `/baseline11` | GET/POST | Sum query params (+ body for POST) |
| `/baseline2` | GET | Sum query params (H2) |
| `/json` | GET | 50-item JSON with computed totals |
| `/compression` | GET | Gzip-compressed large JSON |
| `/db` | GET | SQLite range query |
| `/upload` | POST | Returns byte count |

### Tests subscribed
baseline, pipelined, noisy, limited-conn, json, upload, compression, mixed

cc @carltongibson @felixxm @sarahboyce — thought it'd be cool to see how Django stacks up in HttpArena! Flask vs FastAPI vs Django is the Python comparison everyone's been waiting for.

### DIFF
--- a/frameworks/django/settings.py
+++ b/frameworks/django/settings.py
@@ -6,3 +6,4 @@ MIDDLEWARE = []
 ROOT_URLCONF = 'urls'
 USE_TZ = False
 LOGGING_CONFIG = None
+DATA_UPLOAD_MAX_MEMORY_SIZE = None  # No limit — benchmark uploads 20MB
--- a/frameworks/django/views.py
+++ b/frameworks/django/views.py
@@ -146,7 +146,25 @@ def db_endpoint(request):
 
 @require_http_methods(["POST"])
 def upload_endpoint(request):
-    data = request.body
-    resp = HttpResponse(str(len(data)), content_type='text/plain')
+    # Stream from wsgi.input directly to avoid buffering the entire body
+    # Also handles chunked Transfer-Encoding where CONTENT_LENGTH is absent
+    content_length = request.META.get('CONTENT_LENGTH')
+    stream = request.META['wsgi.input']
+    total = 0
+    if content_length:
+        remaining = int(content_length)
+        while remaining > 0:
+            chunk = stream.read(min(65536, remaining))
+            if not chunk:
+                break
+            total += len(chunk)
+            remaining -= len(chunk)
+    else:
+        while True:
+            chunk = stream.read(65536)
+            if not chunk:
+                break
+            total += len(chunk)
+    resp = HttpResponse(str(total), content_type='text/plain')
     resp['Server'] = 'django'
     return resp


BennyFranciscus:
Django results are in — and honestly, pretty respectable for a full-featured framework running on Gunicorn:

**Highlights:**
- **113K baseline** at 512c — not bad at all for Django with all its middleware stack. Gunicorn workers are doing solid work here
- **121K pipelined** — only a ~7% bump from pipelining, which makes sense since Gunicorn is process-based and each worker handles one request at a time
- **115K JSON** — serialization performance is close to baseline, so the JSON overhead is minimal
- **Memory: 3.9GB baseline** — this is the multi-process model showing up. Each Gunicorn worker is a full Python process with its own memory space

**Interesting patterns:**
- Steep drop at 16384c (113K → 36K) — that's the process-based concurrency model hitting its limits. With a fixed worker pool, high connection counts mean lots of waiting
- Upload returning 4xx across all runs — looks like the upload endpoint config needs a look. Django's CSRF middleware and file upload handling probably need specific configuration for the benchmark's upload format
- Limited-conn profile (113K at 512c, 109K at 4096c) shows Django handles connection reuse efficiently — almost no degradation when connections are reused

**Context:**
For comparison, Flask (also Python/WSGI) hits ~92K baseline. Django coming in 23% higher is interesting — could be the optimized ORM-less path or just Gunicorn tuning differences. Either way, Django's "batteries included" philosophy doesn't cost as much throughput as people assume.

The upload issue should probably be fixed before merging — want me to take a look at the Django config?

PR #85, opened 2026-03-23
Add vif: OCaml 5 web framework with Miou multicore scheduler

BennyFranciscus:
Replaces #84 (closed due to branch protection blocking fixes).

## What

Adds **vif** — an OCaml 5 web framework built on the Miou multicore scheduler, replacing the Dream entry per discussion on PR #25.

### Stack
- **vif** 0.0.1~beta2 + **httpcats** (HTTP/1.1 + HTTP/2)
- **Miou** multicore scheduler (OCaml 5 domains/effects)
- **Yojson** for JSON serialization
- **sqlite3-ocaml** for database queries
- **decompress** (pure OCaml gzip via vif's built-in `~compression:\`Gzip`)

### Changes from #84
- Fixed type error: `Uri.any` for query wildcards vs `Type.any` for POST content-type matching
- Used `nil` for routes that don't accept query parameters (pipeline, json, compression, upload)

### Endpoints
All 8 standard HttpArena endpoints implemented:
- `/pipeline` — simple text response
- `/baseline11` GET/POST — query param sum + body
- `/json` — dataset processing
- `/compression` — gzip compressed response
- `/upload` — streaming body read with byte counting
- `/db` — SQLite query with parameterized range


jerrythetruckdriver:
`json_endpoint` calls `process_items` on every request while `compression` uses the pre-computed `large_payload`. You already solved this problem 10 lines above — just cache the small dataset result at startup too. Third time I've seen this pattern across your PRs.

BennyFranciscus:
Oops — the upload streaming fix I pushed in dce05bf had a typo: `Flux.Source.Source` instead of `Flux.Source`. Fixed in c91f24d, CI should be green now.

Still waiting on the re-benchmark with multicore domains enabled — the first run was only using 1 core (~550% CPU). With the `~domains` fix it should saturate all available cores.

PR #98, opened 2026-03-24
Update nginx to v1.28.3 last stable

joanhey:


BennyFranciscus (approved):
Clean bump — nice touch parameterizing the version in the COPY path too, that hardcoded `1.26.2` in the runtime stage was a landmine waiting to happen. LGTM 👍

joanhey:
I'll add a PR with ngx-php.
But now I'm familiarizing with the bench, and both need to use the same nginx version.

Later will come Workerman.

MDA2AV:
> I'll add a PR with ngx-php. But now I'm familiarizing with the bench, and both need to use the same nginx version.
> 
> Later will come Workerman.

To add a new framework you can either check how to do it manually @ the website Knowledge Base or open an issue and tag @BennyFranciscus  - this is an AI agent with instructions how to add new frameworks and then you only need to polish what he creates by giving him instructions on the PR he creates

PR #99, opened 2026-03-24
fix(nginx,h2o): reject unknown HTTP methods with 405

BennyFranciscus:
As requested in #98 — fixes the `bad method` validation failure for both nginx and h2o.

## Problem

Both the nginx module and h2o accept any HTTP method (`GETT`, `FOOBAR`, etc.) and return 200, failing the `validate.sh` bad method check:
```
FAIL [bad method]: expected 4xx, got HTTP 200
```

### nginx (`ngx_http_httparena_module.c`)
- Method guard at the top of the handler: only `GET`/`HEAD`/`POST` are allowed, everything else gets `405 Method Not Allowed`
- Changed the final fallthrough from `NGX_DECLINED` (which passed to nginx default → 200) to a proper `404 Not Found` for unknown paths

### h2o (`src/main.c`)
- Added `reject_bad_method()` helper that checks for `GET`/`HEAD`/`POST`
- Applied it to all route handlers: pipeline, baseline11, baseline2, json, static, compression, db
- Upload handler already enforced POST-only so it was fine

Both return `405` for unknown methods, which satisfies the validate.sh `4xx` check.

### DIFF
--- a/frameworks/nginx-openresty/nginx.conf
+++ b/frameworks/nginx-openresty/nginx.conf
@@ -35,6 +35,17 @@ http {
     server {
         listen 8080 reuseport;
 
+        # Reject unknown HTTP methods — only allow GET, HEAD, POST
+        access_by_lua_block {
+            local m = ngx.req.get_method()
+            if m ~= "GET" and m ~= "HEAD" and m ~= "POST" then
+                ngx.status = 405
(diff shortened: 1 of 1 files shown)


joanhey:
@BennyFranciscus fix the validation for nginx-openresty, have the same problem.

PR #101, opened 2026-03-25
Add ngx-php

joanhey:
Still need more endpoints. Initial commit.

The json and compression endpoints docs don't specify if the  count can be hard coded.
If need to use a `count()`, we need to add tests to avoid to be hard coded.

The db endpoint documentation don't exist.

PD: first PHP framework.

### DIFF
--- a/frameworks/ngx-php/app.php
+++ b/frameworks/ngx-php/app.php
@@ -14,6 +14,15 @@ $jsonData = json_decode(file_get_contents('/data/dataset.json'), true);
 
 //$bad = fn($x) => !in_array($x, ['POST', 'GET', 'HEAD']);
 
+function guard()
+{
+    if (!in_array(ngx_request_method(), ['POST', 'GET'])) {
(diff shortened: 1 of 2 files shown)


joanhey:
Hi !!

For `keepalive_requests` and `worker_connections` are used the same values than similar nginx in this repo.
I use the same config that:
- nginx https://github.com/MDA2AV/HttpArena/blob/main/frameworks/nginx/nginx.conf
- openresty https://github.com/MDA2AV/HttpArena/blob/main/frameworks/nginx-openresty/nginx.conf

About engine or framework, it's the same than h2o that is an engine and h2o-mruby is marked as framework.
But any app code to run in ngx-php will be writen in PHP, not in C.
But for me it's OK any decision from the maintainers.

PD: also exist h2o with php (but not embedded) as fcgi.

joanhey:
Yes you can merge it.
When I have more time, I'll add more tests.

Thank you !!

PR #168, opened 2026-03-26
[python] FastPySGI: Add ASGI server variant for comparison with WSGI

remittor:
Added ASGI server variant for comparison with WSGI.
In the second commit I will add the implementation for test `async-db`.

### DIFF
--- a/frameworks/fastpysgi-asgi/app.py
+++ b/frameworks/fastpysgi-asgi/app.py
@@ -3,6 +3,7 @@ import sys
 import asyncio
 import json
 import threading
+import multiprocessing
 import zlib
 import sqlite3
 from urllib.parse import parse_qs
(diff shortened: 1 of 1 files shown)


BennyFranciscus (commented):
Nice clean ASGI variant! Code looks solid — couple notes:

- JSON endpoint does per-request computation ✅ and compression does per-request gzip ✅ (large payload pre-serialized but that's fine since the gzip work happens each request)
- Upload streaming looks correct — chunked `receive()` loop without accumulation, just counting bytes
- DB endpoint uses thread-local connections with `PRAGMA mmap_size` — smart for read-heavy workloads
- The forking approach (`os.fork()` for cpu_count workers) is the same pattern as the WSGI variant, should scale nicely

One minor thing: the route handlers accept a `send` parameter that's never used (the main `app` function handles the ASGI send protocol). Not a functional issue at all, just a style thing — could simplify the signatures to `async def pipeline(scope, receive)` etc. if you want.

Excited to see the async-db implementation in the next commit! 🚀

BennyFranciscus:
Good detective work! So `lifespan_startup` is True, meaning the server startup didn't crash — but that doesn't necessarily mean the pool was created successfully. The pool creation could be silently failing and the endpoint might be catching the exception and returning JSON with an empty result set (which would still pass Content-Type validation).

Can you check the logs for something like `DB pool created` or an explicit error from `asyncpg.create_pool`? If the pool is None/unset when the endpoint runs, it'd explain the ~1M req/s — it's just skipping the query entirely.

Re: `127.0.0.1` — yeah, it's worth trying. The CI runner uses a sidecar PostgreSQL container, so the DB host depends on how networking is set up. Check what other frameworks in the arena use for the DB host — look at `run.sh` or the Docker compose setup. It might be `localhost`, `127.0.0.1`, `postgres`, or an env var like `DB_HOST`. Try matching whatever the working frameworks use.

MDA2AV:
I have this test working with results on website for 32 frameworks so far, we can merge this and I can prob fix this locally faster with access to the docker logs if you don't mind @remittor

PR #169, opened 2026-03-26
Add Workerman

joanhey:
# Workerman

An asynchronous event driven PHP socket framework. Supports HTTP, Websocket, SSL and other custom protocols.

All in pure PHP code.

## Stack

- **Language:** PHP
- **Engine:** Workerman (event)

## Endpoints

|----------|--------|-------------|
| `/pipeline` | GET | Returns `ok` (plain text) |
| `/baseline11` | GET | Sums query parameter values |
| `/baseline11` | POST | Sums query parameters + request body |
| `/json` | GET | Processes 50-item dataset, serializes JSON |
| `/compression` | GET | Gzip-compressed large JSON response |
| `/db` | GET | SQLite range query with JSON response |
| `/upload` | POST | Receives 1 MB body, returns byte count |
| `/static/{filename}` | GET | Serves preloaded static files |

## Notes

- No chunked requests for now (WIP)
- Chunked responses OK
- Per-worker database connection

### DIFF
--- a/frameworks/workerman/server.php
+++ b/frameworks/workerman/server.php
@@ -57,9 +57,9 @@ $http_worker->onMessage = static function ($connection, $request) {
             $connection->headers = ['Content-Type' => 'application/json'];
             return $connection->send(json_encode(['items' => $total, 'count' => count($total)]));
         
-        case '/upload':
-            $connection->headers = ['Content-Type' => 'text/plain'];
-            return $connection->send(strlen($request->rawBody()));
+        // case '/upload':
+        //     $connection->headers = ['Content-Type' => 'text/plain'];
+        //     return $connection->send(strlen($request->rawBody()));
 
         case '/compression':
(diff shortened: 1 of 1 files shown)


BennyFranciscus:
You're right, my bad — @MDA2AV already pointed that out to me. Noisy test is designed to send malformed requests so 4xx is expected behavior. I'll stop flagging it 👍

p8:
> A benchmark is not a competition, it's a really good tool to find bugs and improve the code for performance. And I never seen a bench test for chunked content, to improve the performance. But need to be a standalone test, to see the differences in performance.

@joanhey 👋 You are trying to convince a LLM.

joanhey:
I removed the Upload test, because it's the only using chunked body (I think).
In the leaderboard ntex variants also don't have the upload test and others don't have the noisy test.

Without the upload test, we can run the rest of tests.

MDA2AV (approved):

PR #181, opened 2026-03-26
Rails: Decrease threads to 3

p8:
Use an env variable to set it.


BennyFranciscus (approved):
Clean refactor — using `ENV.fetch` makes the thread count configurable without touching the code, and bumping from 4→5 gives each worker an extra thread for handling concurrent requests.

Curious to see if the extra thread per worker helps with the mixed test especially (where DB wait time means threads can overlap more). With PR #174's MN threads enabled, the thread scheduling should handle this well.

One small note: `ENV.fetch('RAILS_MAX_THREADS')` will raise `KeyError` if the env var isn't set (no default). That's fine since you're setting it in the Dockerfile, but if someone runs puma outside Docker they'd need to set it manually. Could do `ENV.fetch('RAILS_MAX_THREADS', '5').to_i` for a safe default — totally optional though since this is a benchmark container.

PR #185, opened 2026-03-26
Rails: use bundler to require gems

p8:


BennyFranciscus (approved):
Same approach as #184 — nice and consistent. Bundler picks up rails, puma, sqlite3 from the Gemfile, so the explicit requires in the controller were redundant. Zlib stays as stdlib.

LGTM 👍

PR #186, opened 2026-03-26
Roda: add framework

p8:


BennyFranciscus:
Hey @p8, noticed the latest push still has `'server' => 'sinatra'` in the `MethodGuard` middleware in config.ru — should be `'server' => 'roda'`. Small thing but it'll show up in the 405 response headers.

Everything else in the rewrite looks clean 👍

BennyFranciscus:
Fix looks good 👍 Server header is correct now. Ready for merge from my side.

PR #204, opened 2026-03-27
Rage: Add framework

p8:


BennyFranciscus (changes requested):
Hey @p8, nice to see Rage getting an entry! Fiber-based non-blocking I/O on Ruby is a cool approach. A few things I spotted:

**Missing files:**
- `config/puma.rb` — Dockerfile CMD references `puma -C config/puma.rb` but this file isn't in the PR. Without it, Puma will use defaults which means it won't bind to `0.0.0.0:8080`. Needs at minimum:
```ruby
bind 'tcp://0.0.0.0:8080'
threads ENV.fetch('RAILS_MAX_THREADS', 3).to_i, ENV.fetch('RAILS_MAX_THREADS', 3).to_i
workers ENV.fetch('WEB_CONCURRENCY', 0).to_i
preload_app!
```
- `config/environment.rb` — `config.ru` does `require_relative 'config/environment'` but there's no `config/environment.rb` in the PR. This will crash on startup.

**Missing gems in Gemfile:**
- `sqlite3` — controller uses `SQLite3::Database` for the db endpoint
- `puma` — used as the app server but not in the Gemfile
- `json` — used for `JSON.parse`/`JSON.generate` (might come bundled with Ruby 4.0 stdlib, but explicit is safer)

**Potential issue with `mattr_accessor`:**
- That's an ActiveSupport method. Does rage-rb provide it? If not, this will blow up at class load time. Regular `class_attribute` or just plain `@@var` with accessor methods might be safer.

**Compression endpoint:**
- `send_data` with `disposition: :inline` — worth checking if Rage supports this. If it's a Rails-specific helper, you might need `render plain: sio.string` with manual headers instead.

**One more thing** — no `static-h2` in the tests list in `meta.json`, which is fine, but just confirming that's intentional (no TLS/H2 support in this config).

Otherwise the controller logic looks solid — same patterns as the Rails entry. Looking forward to seeing how Rage compares!

BennyFranciscus (commented):
Latest push still has the same two things:

1. **Missing `dataset_items` accessor** — line 27 area has `def self.db_available` and `def self.large_json_payload` but no `def self.dataset_items`. The json endpoint calls `self.class.dataset_items` on line 54 → returns `nil` → 500 every time. One-liner fix:
```ruby
def self.dataset_items = @dataset_items
```

2. **`meta.json` engine** — still says `"puma"`. Rage uses Iodine under the hood, so `"iodine"` or just `"rage"` would be more accurate.

Both are tiny fixes, happy to approve once they're in 👍

BennyFranciscus (approved):
LGTM! Both issues resolved — `dataset_items` accessor is in and `meta.json` correctly shows `"engine": "iodine"`. Controller logic looks solid, Rage entry is ready to benchmark 🚀

PR #233, opened 2026-03-28
Add Scotty: Haskell web framework on Warp (first Haskell entry!)

BennyFranciscus:
## Scotty

Adds [Scotty](https://github.com/scotty-web/scotty) — a lightweight Haskell web framework inspired by Ruby's Sinatra, running on the [Warp](https://hackage.haskell.org/package/warp) HTTP server.

**This is the first Haskell entry in HttpArena!** 🎉

### Details

| | |
|---|---|
| **Language** | Haskell (GHC 9.8) |
| **Framework** | Scotty 0.30 |
| **Engine** | Warp 3.4 |
| **Type** | Framework |

### Subscribed Tests

baseline, pipelined, noisy, limited-conn, json, upload, compression, mixed, async-db, static

### Implementation Notes

- Compiled with `-O2 -threaded -rtsopts` and runtime flags `-N -A64m -I0` for maximum throughput
- Dataset and large compression payload pre-loaded into memory at startup
- Static files cached in a `Map` at startup with correct MIME types
- Manual gzip/deflate compression using zlib (compression level 1 for speed)
- SQLite via `sqlite-simple`, PostgreSQL via `postgresql-simple` (connection-per-request with bracket)
- Multi-stage Docker build: haskell:9.8-slim builder → debian:bookworm-slim runtime

### Validation

All 29 validation checks pass locally ✅

cc @scotty-web — would love to see Scotty's numbers on the leaderboard!

/validate


BennyFranciscus:
Pushed a fix — limited RTS to `-N4 -qn4` instead of `-N` (all cores). GHC's thread scheduler creates one capability per core with `-N`, but for IO-bound network work, all that buys you is contention. 4 capabilities with 4 GC threads should be the sweet spot.

If throughput is still low after re-benchmark, could try `-N8` or experiment with the newer threaded IO manager (`-with-rtsopts=--io-manager=native`) available in GHC 9.8+.

Kaliumhexacyanoferrat:
@BennyFranciscus I would not adjust anything regarding the upload test - it is not our job to mask a flaw in the framework implementation. Results should be transparent on this matter. My main concern here is the low CPU saturation which we cannot increase - any ideas left?

@MDA2AV Did you try the upload test with 5 or 10 MB? I have the feeling that this would show a way bigger spread between "good" and "bad" frameworks 🤔

PR #235, opened 2026-03-28
Add Helidon SE 4.4 (Java — Níma virtual threads web server)

BennyFranciscus:
## Helidon SE 4.4.0

Adds [Helidon SE](https://github.com/helidon-io/helidon) — Oracle's cloud-native Java framework with its own **Níma WebServer** built entirely on Java 21 virtual threads.

### Why Helidon?

HttpArena already has several JVM frameworks but they all use either Netty or Tomcat as their underlying server:
- **Spring Boot** → Tomcat
- **Quarkus** → Vert.x/Netty
- **Vert.x** → Netty
- **Ktor** → Netty

Helidon SE 4 is the only JVM entry with its own **custom web server (Níma)** that uses a thread-per-request model on virtual threads instead of event loops. This makes it a genuinely different architecture to benchmark against the reactive/event-loop entries.

### Implementation

- **Framework:** Helidon SE 4.4.0 (WebServer)
- **Engine:** Níma (virtual threads, no Netty)
- **JVM:** Eclipse Temurin 25, ZGC
- **JSON:** Jackson (consistent with other JVM entries)
- **SQLite:** JDBC with thread-local connections, read-only + mmap
- **PostgreSQL:** HikariCP connection pool (64 max, 16 idle)
- **Compression:** Manual gzip with BEST_SPEED level

### Endpoints

| Test | Endpoint | Method |
|------|----------|--------|
| baseline | /baseline11 | GET, POST |
| baseline-h2 | /baseline2 | GET |
| pipelined | /pipeline | GET |
| json | /json | GET |
| upload | /upload | POST |
| compression | /compression | GET |
| db | /db | GET |
| async-db | /async-db | GET |
| static | /static/{filename} | GET |

### Validation

- Docker build: ✅ Clean build with multi-stage (maven → temurin JRE)
- Server startup: ✅ ~18ms (Helidon 4.4.0 with virtual threads)
- All endpoints tested and returning correct responses

### Notes

Helidon 4 rewrote the SE API from async/reactive to blocking (enabled by virtual threads). The result is simpler code that still gets high throughput — will be interesting to see how it stacks up against the Netty-based frameworks.

cc @tomas-langer @spericas (Helidon maintainers)

### DIFF
--- a/frameworks/helidon/Dockerfile
+++ b/frameworks/helidon/Dockerfile
@@ -15,6 +15,5 @@ ENTRYPOINT ["java", \
   "-XX:+UseZGC", \
   "-XX:+UseNUMA", \
   "-XX:+AlwaysPreTouch", \
-  "-XX:-StackTraceInThrowable", \
   "-cp", "app.jar:libs/*", \
   "com.httparena.Main"]


jerrythetruckdriver (commented):
`-XX:-StackTraceInThrowable` disables stack trace generation on all exceptions. No production deployment guide recommends this — it trades debuggability for speed and falls under "experimental or unstable options that trade safety for speed" ([implementation rules](https://www.http-arena.com/docs/add-framework/implementation-rules/)).

Drop that flag. The other JVM flags (ZGC, UseNUMA, AlwaysPreTouch) are standard production tuning and fine to keep.

BennyFranciscus:
Good call — dropped `-XX:-StackTraceInThrowable`. The other flags (ZGC, UseNUMA, AlwaysPreTouch) are standard production tuning so those stay.

PR #237, opened 2026-03-28
Add libreactor (C) — epoll engine entry

BennyFranciscus:
## libreactor

Adds [libreactor](https://github.com/fredrikwidlund/libreactor) v3.0.0 as an **engine** entry.

libreactor is a high-performance C event-driven library by @fredrikwidlund, built around epoll with picohttpparser for HTTP parsing. It's consistently one of the top performers on TechEmpower benchmarks — minimal abstraction, raw C, zero-copy patterns.

### Implementation

- **Type**: Engine (raw HTTP handling, no framework abstractions)
- **Language**: C
- **Event loop**: epoll via libreactor's reactor core
- **HTTP parsing**: picohttpparser (bundled in libreactor)
- **Build**: Compiled from source with `-O3 -march=native -flto`
- **Tests**: baseline, pipelined, limited-conn

### Endpoints

|---|---|---|
| `/baseline11` | GET | Sum query params `a` + `b` |
| `/baseline11` | POST | Sum query params + body (Content-Length and chunked) |
| `/pipeline` | GET | Returns `ok` as `text/plain` |

### Validation

```
=== Results: 7 passed, 0 failed ===
```

All baseline, POST (Content-Length + chunked), anti-cheat randomized inputs, and pipeline checks pass.

cc @fredrikwidlund — would love to see how libreactor performs on HttpArena's benchmarks!

### DIFF
--- a/frameworks/libreactor/Dockerfile
+++ b/frameworks/libreactor/Dockerfile
@@ -1,0 +1,31 @@
+FROM ubuntu:24.04 AS build
+
+RUN apt-get update && apt-get install -y \
+    gcc make autoconf automake libtool pkg-config \
+    libssl-dev git ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+# Build libreactor from source (v3.0.0)
+RUN git clone --depth 1 https://github.com/fredrikwidlund/libreactor.git /tmp/libreactor \
+    && cd /tmp/libreactor \
+    && ./autogen.sh \
+    && ./configure \
+    && make -j$(nproc) \
+    && make install \
+    && ldconfig \
+    && rm -rf /tmp/libreactor
+
+WORKDIR /app
+COPY server.c .
+
+RUN gcc -O3 -march=native -flto -o server server.c \
+    $(pkg-config --cflags --libs libreactor) -lssl -lcrypto
+
+FROM ubuntu:24.04
+RUN apt-get update && apt-get install -y libssl3t64 && rm -rf /var/lib/apt/lists/*
+COPY --from=build /usr/lib/libreactor* /usr/lib/
+COPY --from=build /app/server /server
+RUN ldconfig
+
+EXPOSE 8080
+CMD ["/server"]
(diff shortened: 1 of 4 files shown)


MDA2AV:
increase to 64 reactors or workers, seems to be set to 1

BennyFranciscus:
Good catch — was running a single reactor loop. Pushed a fix: now forks one worker per CPU core, each pinned via `sched_setaffinity`. libreactor's `net_socket` already sets `SO_REUSEPORT`, so each child binds independently.

Should saturate all 64 cores now.

MDA2AV:
set cpus to 64

MDA2AV:
set the value to exactly 64 and dont pin them with sqaffinity

BennyFranciscus:
Done — hardcoded 64 workers and removed `sched_setaffinity` pinning.

MDA2AV:
> Hi, I'm the author. Nice with a new benchmark. This is a somewhat old version of libreactor, I can perhaps submit the new version if that is interesting, which is based on io_uring rather than epoll. Is it correct that strategies to distribute load efficiently is considered against the rules (affinity etc)?

Hi, it is allowed, i asked to disable because it can result in worse performance overall, just to check. 

Would be great if you could submit with the newer version, as it is classified as an engine only the baseline, pipelined and short-lived tests count to the score. The baseline test is locked to 64 cpu threads (cpu has 128, 64 phy cores and 128 SMT), other tests aren't but will likely be in the future for engines to make it simpler to set the reactors/workers/cpu counts

fredrikwidlund:
> Would be great if you could submit with the newer version, as it is classified as an engine only the baseline, pipelined and short-lived tests count to the score. The baseline test is locked to 64 cpu threads (cpu has 128, 64 phy cores and 128 SMT), other tests aren't but will likely be in the future for engines to make it simpler to set the reactors/workers/cpu counts

What is required to classify as production-grade?

BennyFranciscus:
@fredrikwidlund Hey, awesome to have the author here! 🎉 Thanks for checking it out.

I submitted this based on the v2 release on GitHub — would love to see an io_uring-based version from you directly. If you want to open a fresh PR with the newer version, that would be the way to go. You know the library best so you'd be able to tune it properly.

On affinity: yeah, MDA2AV asked me to remove it to test the baseline without pinning. Sounds like it's allowed though if it helps, so feel free to use it in your submission.

Looking forward to seeing the updated version!

PR #240, opened 2026-03-29
[Java] Quarkus: Switch to production recommended APIs and settings

Kaliumhexacyanoferrat:


Kaliumhexacyanoferrat:
@BennyFranciscus any idea why the `/upload` call is failing? Is this a new validation so this never passed in the first place? any changes in the settings that may have caused this?

PR #253, opened 2026-03-29
Allow Content-Encoding header to have no whitespace

p8:
Some framework return headers without whitespace:

    content-type:application/json
    content-encoding:gzip

The `check_header` method already takes this into account, but the encoding header check does not.

### DIFF
--- a/scripts/validate.sh
+++ b/scripts/validate.sh
@@ -318,7 +318,7 @@ if has_test "compression"; then
 
     # Must return Content-Encoding: gzip when Accept-Encoding: gzip is sent
     comp_headers=$(curl -s -D- -o /dev/null -H "Accept-Encoding: gzip" "http://localhost:$PORT/compression")
-    comp_encoding=$(echo "$comp_headers" | grep -i "^content-encoding:" | tr -d '\r' | awk '{print tolower($2)}' || true)
+    comp_encoding=$(echo "$comp_headers" | grep -i "^content-encoding:" | sed 's/^[^:]*: *//' | tr -d '\r' | awk '{print tolower($1)}' || true)
     if [ "$comp_encoding" = "gzip" ]; then
         echo "  PASS [compression Content-Encoding: gzip]"
         PASS=$((PASS + 1))


BennyFranciscus (commented):
Hey p8, good catch on the whitespace-less headers! 👍

One issue though — the fix still has `awk '{print tolower($2)}'` at the end of the pipe, but after the `sed 's/^[^:]*: *//'` strips the header name + colon, the remaining value (e.g. "gzip") becomes `$1` not `$2`.

So with the current fix:
- `"content-encoding: gzip"` → sed → `"gzip"` → awk `$2` = empty → FAIL
- `"content-encoding:gzip"` → sed → `"gzip"` → awk `$2` = empty → FAIL

Should be `$1` instead of `$2`:
```bash
comp_encoding=$(echo "$comp_headers" | grep -i "^content-encoding:" | sed 's/^[^:]*:[[:space:]]*//' | tr -d '\\r' | awk '{print tolower($1)}' || true)
```

Or simpler — keep the awk-only approach but make it handle both formats:
```bash
comp_encoding=$(echo "$comp_headers" | grep -i "^content-encoding:" | tr -d '\\r' | awk -F: '{gsub(/^[[:space:]]+/, "", $2); print tolower($2)}' || true)
```

This uses `:` as the field separator so `$2` is everything after the colon, then strips leading whitespace.

BennyFranciscus:
Hey @p8, the intent here is right but I think there's a bug in the sed+awk pipeline.

After `sed 's/^[^:]*: *//'` strips the header name, the value becomes `$1` for awk, not `$2`:

```
# Input: "content-encoding:gzip"
# After sed: "gzip"
# awk {print $2} → empty (gzip is $1)

# Input: "content-encoding: gzip"  
# After sed: "gzip"
# awk {print $2} → empty (same issue)
```

Simplest fix — either change `$2` to `$1`:
```bash
comp_encoding=$(echo "$comp_headers" | grep -i "^content-encoding:" | sed 's/^[^:]*: *//' | tr -d '\\r' | awk '{print tolower($1)}' || true)
```

Or skip sed entirely and use a more flexible awk delimiter:
```bash
comp_encoding=$(echo "$comp_headers" | grep -i "^content-encoding:" | tr -d '\\r' | awk -F ":[[:space:]]*" '{print tolower($2)}' || true)
```

The second approach handles both `content-encoding: gzip` and `content-encoding:gzip` in one step.

BennyFranciscus (approved):
Looks good now 👍 The sed+awk pipe handles both cases cleanly.

MDA2AV (approved):

PR #260, opened 2026-03-29
 Add Fletch framework implementation for HttpArena (H1 + mixed + async-db)

kartikey321:
This PR adds a new fletch framework entry under frameworks/fletch with
  a Dockerized Dart server implementation for HttpArena profiles.

  What’s included

  - New framework directory:
      - frameworks/fletch/meta.json
      - frameworks/fletch/Dockerfile
      - frameworks/fletch/pubspec.yaml
      - frameworks/fletch/pubspec.lock
      - frameworks/fletch/bin/server.dart
  - Implemented endpoints for subscribed profiles:
      - /baseline11 (GET/POST)
      - /pipeline
      - /json
      - /upload
      - /compression
      - /static/:filename
      - /db
      - /async-db
  - async-db uses DATABASE_URL and returns the required JSON shape.
  - Added lazy/retry async-db connection behavior so startup DB
    unavailability can recover on subsequent requests.
  - Updated dependencies to latest compatible versions (notably sqlite3
    ^3.2.0).
  - Updated Docker build flow to support sqlite3 3.x build hooks:
      - uses dart build cli
      - runs bundled executable from /server/bin/server
  - Runtime image includes libsqlite3-dev to provide libsqlite3.so
    required by Dart sqlite FFI.


MDA2AV:
Hey @kartikey321 welcome to the arena!

kartikey321:
@BennyFranciscus @MDA2AV 
Thanks for the catch. I double-checked this in the current branch:

  Fletch already scales worker count by CPU visibility in the container:
  final n = Platform.numberOfProcessors;
  then it spawns n - 1 isolates + runs one on the main isolate.

  So isolate count is not hardcoded to a small number.

  Given the benchmark logs, CPU around ~1200–1400% could also be due to
  runner/container limits or loadgen saturation (benchmark script uses
  THREADS=64 for gcannon by default), not necessarily isolate under-
  spawning.

  Could we confirm runner/container CPU visibility for this run?

  - nproc inside the benchmark container
  - docker info --format '{{.NCPU}}' on the runner

  If helpful, I can also test a rerun with higher loadgen thread count
  (e.g. THREADS=128) to see whether CPU utilization and throughput scale
  further.

BennyFranciscus:
That's a really solid detective job tracing it to `Platform.numberOfProcessors` vs cgroup limits — and yeah, every other framework in the benchmark hits the same thing and works around it with `nproc`.

The fact that it *didn't* improve is actually the interesting part though. If you're now spawning 64 isolates on 64 CPUs and still only hitting ~1360% CPU, the bottleneck isn't isolate count — it's something upstream preventing those isolates from saturating their cores.

A few things worth investigating:

1. **Accept queue contention** — are you using `HttpServer.bind(..., shared: true)` across isolates? Even with `shared: true`, Dart's implementation goes through a single accept lock internally. At 64 cores that lock can become the bottleneck itself, with most isolates spinning waiting to accept rather than doing work.

2. **Single-listener distribution** — if connections are accepted on one isolate and then handed off, that main isolate becomes the ceiling. Some Dart servers use `SO_REUSEPORT` via FFI to let the kernel distribute across sockets directly, which avoids that entirely.

3. **Event loop saturation per isolate** — Dart isolates are single-threaded event loops. If each isolate can only push ~14K req/s (180K / ~13 active cores), that might just be the ceiling of `dart:io`'s HTTP parser per event loop. Worth comparing against a raw `dart:io` server without Fletch's routing layer to see if the framework adds overhead or if it's the runtime itself.

The profiler idea is probably the fastest path to an answer — specifically tracking where each isolate spends its time (accept vs parse vs handler vs write). That'll tell you immediately if it's contention or per-isolate throughput.

Re: Content-Length — no worries, let's wait for the call on that and iterate.

Want me to kick off another benchmark run after you push changes, or should we merge as-is and iterate from main?

PR #264, opened 2026-03-30
fix(fastify): use native Fastify HTTP/2 instead of raw node:http2

### LINKED ISSUE #261: [Audit] fastify: HTTP/2 endpoints bypass Fastify entirely
## Violation

**Rule:** Use framework-level APIs

**File:** `frameworks/fastify/server.js`

The HTTP/2 server on port 8443 (`startH2()` function, ~line 195) uses raw `node:http2` with a hand-rolled request handler:

```js
const h2server = http2.createSecureServer(opts, (req, res) => {
    // manual URL parsing, manual query string parsing, manual response writing
    // no Fastify involved at all
});
```

Fastify natively supports HTTP/2 — you just pass `http2: true` to the constructor:

```js
const app = Fastify({ http2: true, https: { cert, key } });
```

The `meta.json` declares `baseline-h2` and `static-h2` tests, so these endpoints are being benchmarked. But they measure raw Node.js HTTP/2 performance, not Fastify.

## Suggested fix

Create the H2 server as a second Fastify instance with `http2: true` and reuse the same route handlers, similar to how the HTTP/1.1 server is set up.

BennyFranciscus:
## What

The H2 handler was bypassing Fastify entirely — using a raw `node:http2` `createSecureServer` callback with manual URL parsing and direct `res.writeHead()`/`res.end()` calls. This means the baseline-h2 and static-h2 tests were measuring Node.js core HTTP/2, not Fastify.

## Fix

Replace with a second Fastify instance configured with `{ http2: true, https: { ... } }`, registering the same route handlers through Fastify's router and middleware pipeline.

The H2 routes are extracted into a shared `registerH2Routes()` function to avoid duplication with the H1 app (baseline and static endpoints are identical).

## Changes

- Remove `require('http2')` — no longer needed
- Add `registerH2Routes()` helper with baseline + static routes
- `startH2()` now creates a proper Fastify instance with `http2: true`
- Content type parsers registered on both H1 and H2 instances

Fixes #261

cc @Kaliumhexacyanoferrat @jerry

### DIFF
--- a/frameworks/fastify/server.js
+++ b/frameworks/fastify/server.js
@@ -91,44 +91,12 @@ function startWorker() {
     app.addContentTypeParser('application/octet-stream', { parseAs: 'buffer' }, (req, body, done) => done(null, body));
     app.addContentTypeParser('*', { parseAs: 'buffer' }, (req, body, done) => done(null, body));
 
-    // --- /static/:filename ---
-    app.get('/static/:filename', (req, reply) => {
-        const sf = staticFiles[req.params.filename];
-        if (sf) {
-            reply
-                .header('server', SERVER_NAME)
-                .header('content-type', sf.ct)
-                .header('content-length', sf.buf.length)
-                .send(sf.buf);
-        } else {
-            reply.code(404).send();
-        }
-    });
+    // Register shared routes (baseline, static)
+    registerSharedRoutes(app);
 
     // --- /pipeline ---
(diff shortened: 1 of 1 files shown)


BennyFranciscus:
Done! Extracted the shared handlers into `registerSharedRoutes()` — both H1 and H2 instances now use the same function for static, baseline, and POST routes. Net -34 lines 🧹

PR #281, opened 2026-03-30
[Rust] actix: Switch to high-level APIs

Kaliumhexacyanoferrat:


Kaliumhexacyanoferrat:
@MDA2AV Ready to be merged, but I would keep this one one tuned. Requires someone who knows how production apps with actex really look like 🙄

PR #290, opened 2026-03-30
Add pgsql to run.sh

joanhey:
With this change I can test the async-db endpoint.

## Description

---

**PR Commands** — comment on this PR to trigger:

| Command | Description |
|---------|-------------|


joanhey:
We need to force that is removed when Ctrl + c

PR #324, opened 2026-04-01
[C#] Switch to Debian release images

Kaliumhexacyanoferrat:
For alpine we need to install the very same version of msquic as the dotnet image does or it will fail to load the `.so` dependency.


MDA2AV:
Running local benchmarks for merged frameworks, Ill trigger validation and benchmarks for this after

PR #369, opened 2026-04-02
Roda: Use prepared statements

p8:
Also simplify code and make it easier to test locally.


p8:
@MDA2AV Ah ok, the noisy tests are showing a lot of errors now. I'll take a look.

PR #377, opened 2026-04-03
Roda: freeze loaded data

p8:


MDA2AV:
We are limiting the CPU cores available for all tests so the comparison is invalid vs the values on website which have unlimited cores

PR #467, opened 2026-04-10
Rage: Reduce connection pool size

p8:
Make the pool size as big as the number of processors. This might fix the issue of 5xx responses if a connection times out if it is not getting processed.


MDA2AV:
Hey @p8 yesterday when doing a full run I noticed 2 of the ruby frameworks had issues, I think there was a merge issue with them, send command for a full benchmark run

p8:
Thanks @MDA2AV !

p8:
This doesn't improve things so closing.

PR #474, opened 2026-04-11
feat(fletch): add Dart HTTP framework with multi-process scaling 

kartikey321:
## Description

Adds the Fletch Dart HTTP framework with a multi-process architecture that mirrors
  Node.js cluster mode.                                                                  
                  
  - Multi-process model: entrypoint.sh spawns N independent OS processes (one per vCPU), 
  each with its own Dart VM EventHandler thread — bypassing Dart's
  single-EventHandler-per-process limitation                                             
  - SO_REUSEPORT shim: reuseport_shim.c is an LD_PRELOAD library that intercepts bind()
  and sets SO_REUSEPORT on TCP sockets, letting the kernel distribute connections evenly 
  across all processes (workaround for dart-lang/sdk#10843)
  - cgroup-aware CPU detection: reads cgroupv2 (cpu.max) and cgroupv1 (cpu.cfs_quota_us) 
  so --cpus=N Docker limits are respected correctly                                      
   
  Architecture                                                                           
                  
  entrypoint.sh
    └── N × /server/bin/server  (one OS process per vCPU)
          └── each: 1 Dart isolate + 1 EventHandler (epoll) thread                       
                └── SO_REUSEPORT via LD_PRELOAD → kernel load-balances connections       
                                                                                         
  Supported profiles                                                                     
                                                                                         
  baseline, pipelined, limited-conn, json, upload, compression, noisy, mixed, static,    
  async-db
                                                                                         
  Test plan       

  - Docker image builds (two-stage: AOT compile + gcc shim)                              
  - Core endpoints verified: /pipeline, /json, /static, /upload
  - Multi-process spawning confirmed: process count matches container CPU quota          
  - 36k req/s on /pipeline in local smoke test (11-core Mac, Docker)                     
  - Full benchmark suite via CI 

---

| Command | Description |
|---------|-------------|

---

<details>

You can validate and benchmark your framework on your own machine using the lite scripts — no CPU pinning, fixed connections, all load generators run in Docker.

**Linux:**
```bash
```

**Windows / macOS (Docker Desktop):**
```bash
./scripts/validate-windows.sh <framework>
./scripts/benchmark-lite-windows.sh <framework> baseline
./scripts/benchmark-lite-windows.sh --load-threads 4 <framework>
```

The `-docker` variants use a Docker bridge network instead of `--network host` (which doesn't work on Docker Desktop).

**Requirements:** Docker Engine. Load generators (gcannon, h2load) are built as Docker images on first run. gcannon source is expected at `../gcannon` (override with `GCANNON_SRC`).

</details>


kartikey321:
Phase 1 — What's in the PR now                                                  
   
  N processes × 1 isolate. Each process = independent EventHandler + independent  
  GC heap. Proves the theory.
                                                                                  
  If the CI benchmark shows linear scaling past the old 1300% plateau → theory    
  confirmed, move to Phase 2.
                                                                                  
  ---             
  Phase 2 — The hybrid
                      
  Once you have real benchmark numbers showing where multi-process peaks (the new
  ceiling), you use that data to find the optimal K:                              
   
  optimal_K = peak_isolate_count_before_eventhandler_saturates / 2                
                                                                                  
  Then instead of N×1, you run:
  ceil(N / K) processes × K isolates each                                         
                                         
  So on a 64-core machine with K=6:                                               
  - 10-11 processes instead of 64                                                 
  - Each EventHandler at ~50% load instead of either saturated (multi-isolate) or 
  idle (pure multi-process)                                                       
  - ~300MB RAM instead of ~1.9GB

kartikey321:
@MDA2AV  @BennyFranciscus  

Is pre-serialising response bytes at worker startup allowed for /json, /compression, and /upload?
I noticed a few bottlenecks I’d like to fix. Before doing so, I wanted to check if the following optimisations are within the rules, since they’re already used by other frameworks in the benchmark.
1. Pre-serialising JSON at worker startup (/json, /compression)
Both Bun and Hono load and serialise the large dataset once per worker at startup:
Dart// bun / hono — top of worker module
const largeItems = largeData.map(d => ({ ...d, total: ... }));
const largeJsonBuf = Buffer.from(JSON.stringify({ 
  items: largeItems, 
  count: largeItems.length 
}));
// then per request:
const compressed = Bun.gzipSync(largeJsonBuf, { level: 1 });
Actix does the same with a json_large_cache: Bytes field in AppState built at startup.
Is it valid to do the same in our Dart isolates — pre-encode the dataset to Uint8List once at _run() startup, then gzip it per request for /compression?

Thanks!

MDA2AV:
For Json test you can load the dataset.json at startup but
On each GET /json request, the server:

    Iterates all 50 items
    Computes a total field (price × quantity) for each item
    Builds the response object
    Serializes everything as JSON (~10 KB response)

For compression test yea, it is not expected that you serialize, just compress every request

Also I would wait a few days because we are changing some tests, adding more complexity to some and removing some like sync-db. The changes will be merged probably until tomorrow end of the day

MDA2AV:
We changed some tests and added some, for that reason validation is now failing because there are some new extra parameters for some tests

kartikey321:
Wow! the amount of effort i put in the profilers to identify the bottlenecks, it didn't event work
Barely moved 2-3%

kartikey321:
Will add json-comp and json-tls support and then this should be ready to merge.

I believe further optimizations at the server level are limited by dart:io's architecture — the HTTP parser is implemented in Dart  and every I/O operation crosses from the C++ EventHandler into the Dart isolate via port messages. Closing that gap would require bypassing dart:io entirely with a native FFI-based HTTP layer, which is out of scope for now, as there isnt a prod level setup now

MDA2AV (approved):

PR #476, opened 2026-04-11
Add swerver framework

justinGrosvenor:
## Summary

- Add **swerver** — a high-performance HTTP server written in Zig
- Native HTTP/1.1 + HTTP/2 (HPACK) + HTTP/3 (QUIC, RFC 9000-9002)
- Single-threaded event loop per worker with SO_REUSEPORT fork fan-out
- Zero-copy request parsing, fixed-size buffer pools, no GC
- kqueue (macOS/BSD), epoll (Linux), io_uring (Linux) backends
- Builds on Debian Trixie (OpenSSL 3.5+ required for QUIC TLS callbacks)

## Test profiles

`baseline`, `pipelined`, `limited-conn`, `noisy`, `tcp-frag`, `upload`, `json`, `static`, `baseline-h2`, `static-h2`, `baseline-h3`, `static-h3`

## Type

`production` — no reverse proxy, no response caching for dynamic endpoints, standard library TLS.

## Container

Two swerver processes in a single container:
- `:8080` — plaintext HTTP/1.1
- `:8443` — TLS HTTP/1.1 + HTTP/2 (ALPN) + QUIC HTTP/3

## Repo

https://github.com/justinGrosvenor/swerver

cc @BennyFranciscus


MDA2AV:
@justinGrosvenor your server is running using 1 CPU thread, probably missing some config

MDA2AV:
Hmm, still low, there are 64 cpus available for the docker, optimal is 6400%

justinGrosvenor:
I see the bottle neck now. will verify before i push again :)

MDA2AV:
Ok, don't use the validate command, I'll trigger validation, github actions are not doing well today

MDA2AV:
Ok, looks like working fine but a few points:

- This is engine type, not production
- I recommend you only implement baseline (h1, h2 and h3), pipelined, limited-con tests because these tests are not going to change and any change to the other tests requires an update directly on the framework repo. It is fine though if you want to keep them, just an heads up though we are changing many tests this weekend so that will require you to update the implementation on json, upload and static

joanhey:
image

MDA2AV:
> image

json endpoint is no longer in the list

joanhey:
If a validate test fail, don't need to run the benchmark for this endpoint.

image

It's the same which test fail, if fail the benchmark never need to run for this endpoint.

MDA2AV (approved):

joanhey:
Sorry, but where is the code for this benchmark ???

frameworks/swerver/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    mkdir -p /opt/zig; \
+    tar -xJf /tmp/zig.tar.xz -C /opt/zig --strip-components=1; \
+    rm /tmp/zig.tar.xz; \
+    ln -s /opt/zig/zig /usr/local/bin/zig; \
+    zig version
+
+# Clone swerver source — pinned to specific commit to bust Docker cache
+WORKDIR /src
+RUN git clone https://github.com/justinGrosvenor/swerver.git . && \
+    git checkout 3ddb613
```
joanhey: Please use a branch or tag!!

Any dev can create a commit only for the bench, but never in the real framework.

joanhey (changes requested):

MDA2AV:
> Sorry, but where is the code for this benchmark ???

It is embedded into the framework itself, that's why marked as engine

joanhey:
But the code for the bench need to be in this repo.
So any dev can check it !!

joanhey:
This code have 2 fails, we need to go to the other repo for mark them as bad.

MDA2AV:
> This code have 2 fails, we need to go to the other repo for mark them as bad.

what fails? json is not enabled

If the framework itself is the benchmark I prefer this, I don't want the whole implementation here, engines are for that, experimental

joanhey:
Yes, but the code for a benchmark need to be always in the same repo.

So devs can compare, help and learn.
If he make any change in his repo than break the requirements, no body will notice it !!

MDA2AV:
> So devs can compare, help and learn. If he make any change in his repo than break the requirements, no body will notice it !!

That only applies to tuned, production and upcoming new category proxy. Engines as I said, no rules.

I already reviewed this framework implementation

justinGrosvenor:
Thanks for the reviews and for letting me optimize on your threadripper. 🔥

Swerver is very young and will be evolving a bit more soon including moving the benchmarking stuff outside the code. We can stay pinned to the current commit here so nothing breaks and revisit in a few weeks when the code and tests have stabilized. Maybe even as "tuned"

MDA2AV:
> Swerver is very young and will be evolving a bit more soon including moving the benchmarking stuff outside the code. We can stay pinned to the current commit here so nothing breaks and revisit in a few weeks when the code and tests have stabilized. Maybe even as "tuned"

Take your time, engine section is exactly for this kind of projects, experimental runtimes and barebones frameworks. Ideally it is better to have the benchmark implementation code here in this repo so when possible I would appreciate it

MDA2AV (approved):

PR #489, opened 2026-04-13
add elysia

SaltyAom:
## Description

add [Elysia](https://elysiajs.com) to benchmark

**Note:
1. I have `ulimit` set to `unlimited`, which broke the validate script (can get over by modifying the `validate.sh`

But I'm not sure if things work as intended
Screenshot 2569-04-13 at 08 16 21

2. I'm on macOS, non of the benchmark scripts work
Screenshot 2569-04-13 at 08 17 51

---

| Command | Description |
|---------|-------------|

---

<details>

You can validate and benchmark your framework on your own machine using the lite scripts — no CPU pinning, fixed connections, all load generators run in Docker.

**Linux:**
```bash
```

**Windows / macOS (Docker Desktop):**
```bash
./scripts/validate-windows.sh <framework>
./scripts/benchmark-lite-windows.sh <framework> baseline
./scripts/benchmark-lite-windows.sh --load-threads 4 <framework>
```

The `-docker` variants use a Docker bridge network instead of `--network host` (which doesn't work on Docker Desktop).

**Requirements:** Docker Engine. Load generators (gcannon, h2load) are built as Docker images on first run. gcannon source is expected at `../gcannon` (override with `GCANNON_SRC`).

</details>


MDA2AV:
> 2. I'm on macOS, non of the benchmark scripts work

Yea, that's a work in progress. We should have fully docker benchmark versions eventually for windows and mac

SaltyAom:
I think this should fixed `POST /upload` validate, probably

MDA2AV:
the benchmark.db no longer exists, we removed that test, it's only postgres now. I can fix this locally probably, probably server crashes trying to load the file

MDA2AV:
Sorry for vandalizing, it was crashing every docker so I reused an existing implementation we had for elysia.

Not perfect but working, I'll eventually check elysia docs to use proper framework api for static files and compression, and fix async-db.

MDA2AV (approved):

yokeTH:
@SaltyAom you added .DS_Store to this repo 🤣

SaltyAom:
> @SaltyAom you added .DS_Store to this repo 🤣

I'm sorry. I have failed you guys 

image

SaltyAom:
> Sorry for vandalizing, it was crashing every docker so I reused an existing implementation we had for elysia.
> 
> Not perfect but working, I'll eventually check elysia docs to use proper framework api for static files and compression, and fix async-db.

How can I help to integrate my implementation into the benchmark?

Because the current implementation is doing more work than it should, and makes Elysia significantly slower than it is

MDA2AV:
> How can I help to integrate my implementation into the benchmark?
> 
> Because the current implementation is doing more work than it should, and makes Elysia significantly slower than it is

Open PR with changes we benchmark it every iteration on the PR until you think it is ready to merge. Note that for production type we need to always use the documented API, no hacks or configs that cannot be found on the official elysia docs. Still those can be applied if we mark it as tuned.

I'd suggest small changes iteratively, we can benchmark 20 times if needed, are there any specific tests you want to improve? The current elysia results are in line of what we have with bun runtime isolated and other frameworks that use bun like hono


If you want to iterate locally then I'd wait a few more days as I am still working on the scripts, and then need to validate on mac and windows

SaltyAom:
> > How can I help to integrate my implementation into the benchmark?
> 
> > 
> 
> > Because the current implementation is doing more work than it should, and makes Elysia significantly slower than it is
> 
> 
> 
> Open PR with changes we benchmark it every iteration on the PR until you think it is ready to merge. Note that for production type we need to always use the documented API, no hacks or configs that cannot be found on the official elysia docs. Still those can be applied if we mark it as tuned.
> 
> 
> 
> I'd suggest small changes iteratively, we can benchmark 20 times if needed, are there any specific tests you want to improve? The current elysia results are in line of what we have with bun runtime isolated and other frameworks that use bun like hono
> 
> 
> 
> 
> 
> If you want to iterate locally then I'd wait a few more days as I am still working on the scripts, and then need to validate on mac and windows

My original implementation would be most ideal if possible, modify if needed

If I were to open another PR, I would use the same one but modify to see what happen but because I'm not sure what happened when you said it was crashing every other docker, I'm not sure what to do

MDA2AV:
> If I were to open another PR, I would use the same one but modify to see what happen but because I'm not sure what happened when you said it was crashing every other docker, I'm not sure what to do

Ok, so just as disclaimer I know almost zero about typescript so I might be wrong in what I'll say here.

I think the issue in the implementation you submitted:
```
for (let i = 0; i < availableParallelism(); i++) {
	if (cluster.isPrimary) cluster.fork();
	else {
		// worker startup
		new Elysia().listen(8080);
	}
}
```
availableParallelism is the docker's cpu quota which is half the processor available quota so 128/2 = 64.

does two different things at once:
 - In the primary, it forks 64 workers.
 - In each worker, it loops 64 times and runs startup 64 times.

You do not end up with 4096 worker processes, you end up with 64 worker processes, but each worker tries to create 64 server instances / app initializations. This can lead to issues trying to connect to postgres 4096 times, and initialize the module way too many times.

So what I suggest is probably something like

```
if (cluster.isPrimary) {
	const workers = availableParallelism();
	for (let i = 0; i < workers; i++) { cluster.fork(); }
} else { await startWorker(); }
```

I think the rest I can try to keep exactly the same, the sqlite test was removed so I'll need to remove the logic for it.

SaltyAom:
Yeah, that seems like it

I didn't sleep that night when I wrote that code

PR #503, opened 2026-04-14
Helidon - Java Microservice Framework, production type

tomas-langer:
Validated benchmark on my machine.
Ran the full benchmark as well to make sure we can pass.

# Project

This framework runs Helidon SE 4.4.1 on Níma WebServer as a `production`
benchmark entry.

The current subscribed benchmark profiles are:

- `baseline`
- `pipelined`
- `limited-conn`
- `json`
- `json-comp`
- `json-tls`
- `upload`
- `static`
- `async-db`
- `api-4`
- `api-16`
- `baseline-h2`
- `static-h2`
- `unary-grpc`
- `unary-grpc-tls`
- `echo-ws`

Profiles not currently supported here:

- HTTP/3: `baseline-h3`, `static-h3`
- `gateway-64`

# Listener layout

The benchmark wiring is split by listener:

- `8080` (`default`): HTTP/1.1 endpoints, cleartext gRPC, and WebSocket
- `8081` (`h1-tls`): HTTP/1.1 + TLS for `json-tls`
- `8443` (`h2-tls`): HTTP/2 + TLS for `baseline-h2`, `static-h2`, and TLS gRPC

Static content and TLS are configured from `application.yaml`, not
programmatically.

# Divergence from benchmark guidance

## `async-db` uses JDBC + HikariCP

The benchmark guidance for `async-db` prefers an async PostgreSQL driver.
This Helidon entry currently uses the standard PostgreSQL JDBC driver with
HikariCP.

That means the implementation is benchmark-contract correct, but it does not
follow the async-driver recommendation literally. This is an intentional
tradeoff for the current Helidon/Níma production entry.

Helidon WebServer is designed for Java Virtual Threads and optimized for blocking operations.


MDA2AV:
About static test - pre compressed files (both .br and .gz) are present/available along with the original files so even for production types if the framework supports it is allowed to serve those directly (when requests allows compressed response) instead of compressing them every request or serving the full uncompressed files

tomas-langer:
> About static test - pre compressed files (both .br and .gz) are present/available along with the original files so even for production types if the framework supports it is allowed to serve those directly (when requests allows compressed response) instead of compressing them every request or serving the full uncompressed files

I may do that as a follow up - our static content does not have a built-in support for something like this, so I would have to create a "manual" handler for static content. Considering this would be a huge performance improvement, I will probably need to do it, as the current numbers look awful (due to the compression happening each time).

MDA2AV:
Ok, I'll investigate the low CPU usage on some tests (json and async-db), probably low Conn numbers, those will be bumped in the future to 16k and 32k for some tests as we gather more data on which numbers can be more fitting. Rest looks good to me, we added grpc stream test too if it is of interest.

MDA2AV (approved):

tomas-langer:
> Ok, I'll investigate the low CPU usage on some tests (json and async-db), probably low Conn numbers, those will be bumped in the future to 16k and 32k for some tests as we gather more data on which numbers can be more fitting. Rest looks good to me, we added grpc stream test too if it is of interest.

We get much better results with `tcp-no-delay` enabled, but I cannot do it for a production test. This seems to be specific to JSON tests (probably a function of how we serialize the data).

remittor:
This is strange:
```
json     | 4096 | 186,515
json-tls | 4096 | 244,830
```

MDA2AV:
> This is strange:
> 
> ```
> json     | 4096 | 186,515
> json-tls | 4096 | 244,830
> ```

Not when you see that json-tls uses 2x the CPU. Probably wont happen with 16k concurrent conn

MDA2AV:
> We get much better results with `tcp-no-delay` enabled, but I cannot do it for a production test. This seems to be specific to JSON tests (probably a function of how we serialize the data).

Yes, for production NoDelay/disable nagle's algo are not permitted unless they are the framework's default parameters, we can add a tuned type for helidon to compare but only production types go into rounds and stay archived

tomas-langer:
I think the reason JSON is faster of SSL is that the real probem is on the socket layer, and SSL socket handling is implicitly different, and probably better for this specific benchmark. We will investigate this in Helidon.

MDA2AV:
One little note here is that in these tests connections are not opened and kept alive during the whole duration, in some cases like json connections only stay alive for 25 requests and then a new one is opened, this can be simulated with [gcannon](https://github.com/MDA2AV/gcannon) -r parameter (requests per connection), we've seen huge impact in some frameworks where creating new connections is more expensive and can idle the CPU a lot.

PR #508, opened 2026-04-14
Helidon - Java Microservice Framework, tuned type

tomas-langer:
Addition of stream grpc to both Helidon versions

## Description
A follow up for #503, this PR adds a Helidon tuned version.


tomas-langer:
sorry, committed too much, I will re-push

done

MDA2AV:
you have 180k loc of logs, not sure if you want to check them

tomas-langer:
I will run the benchmark locally and check the logs. Is it to a specific benchmark?

MDA2AV:
> I will run the benchmark locally and check the logs. Is it to a specific benchmark?

You can see them here in this PR commit, when we do --save they get stored

tomas-langer:
> 🚀 Benchmark run triggered for `helidon-tuned` (all tests) with --save. Results will be posted here when done.

I will analyze the results and push an updated version to get rid of these log files. I hope this is not a bug in the product, as that would take a bit longer time to fix...

tomas-langer:
It seems that a lot of the cases are "peer resets", where we just need to gracefully consume the exception when writing a response.

MDA2AV:
Ah ok, I don't mind the logs. Just pointed as could be deteriorating perf. You can also see docker logs for each test on website, for each detailed test view if you click on the framework row, a popup show and on the bottom should have link to docker logs on that test

MDA2AV (approved):

tomas-langer:
Thanks, I will try to both update the benchmark to be less verbose, and fix this in Helidon directly, so we do not need to do a benchmark fix...

PR #515, opened 2026-04-15
elysia improvements

MDA2AV:
Improve elysia performance


MDA2AV:
@SaltyAom 

Changed back to your original implementation as much as possible, improved the performance for the db and upload tests with better pooling for postgres and reading in chunks for upload. 

So apparently Elysia will automatically set the content type headers for numbers and objects but not for strings (pipelined test case) and setting that header manually lowers throughput from  4.8M to 80K... (what?!) 

Anyway, doesn't matter much as pipelined test is irrelevant and likely to be removed but this may be an issue if any tests needs to return a string

PR #518, opened 2026-04-15
Fix workerman running issues

MDA2AV:


joanhey:
I don't understand it, the Pgsql is the same than with other framewors: ngx-php, ...
This change it's from benny :hand: , and later we'll check it again.

But for now all it's working :+1:

joanhey:
Benny only hide errors, that is what need the developers.

Thanks again.

joanhey:
The only code than I think fix the problem is this:
https://github.com/MDA2AV/HttpArena/pull/518/changes#diff-bffc971fe767323c96a6323344b244eec4d42c1898c7a05d89ea232dcb72e8f5R10-R12

But ngx-php don't have it an have all endpoints working !!

I's only a comment.

MDA2AV:
Yea the issue was around the pg database, I guess you can probably debug it better, I just added that rail and it worked so we can go like this for now and eventually improve if needed

PR #559, opened 2026-04-19
[ngx-php] Remove guard function

joanhey:
## Description

- The guard function in the nginx access point is removed, as the new crud endpoint need all methods.
  And nginx return a 400 Bad method for any incorrect method, before arrive to the locations.

- Add backlog to nginx listen, as the server have set `somaxconn 65535`.

- Fix Post problem in the baseline, thanks to checking the docker logs that are in this repo.
- Still try to fix the upload problem, but locally we don't have logs to check the errors.

### Local platform problems
It's very difficult to test and optimize any framework locally.

- `Validate.sh `don't use the docker build cache, so each time need to download and build the image. It's very slow for each validate, and also we can top the max reqs to docker and github.
  With `run.sh` the build is very fast, almost instantly. 
- The `benchmark-lite.sh` still don't work, fail to build the load generators images.
- We don't have logs to check any error. With `Validate.sh` showing the logs when finish, it'll be enough.
- When `benchmark-lite.sh` work, how we can test the json comp score formula ?
  It's a very good reference for any framework to balance size and performance.
  

---

| Command | Description |
|---------|-------------|

---

<details>

```bash
```

</details>


MDA2AV:
Can you change this to tuned?

joanhey:
Why ?

joanhey:
If you explain why we need to change it, we can decide about any change that is breaking the production, to delete it or use tuned.

MDA2AV:
worker_cpu_affinity auto;
worker_rlimit_nofile 65536;
timer_resolution 1s;
worker_connections 65536;
keepalive_requests 1000000;
access_log off;
etag off;
gzip_static on;
brotli on;
brotli_static on;
listen ... reuseport backlog=65536;
php_ini_path ...;
init_worker_by_php ...;

These are not default nginx, can you find anywhere in nginx docs recommending to set these as default configs?

joanhey:
In the nginx documentation  https://nginx.org/en/docs/.

**Core functionality** https://nginx.org/en/docs/ngx_core_module.html
- https://nginx.org/en/docs/ngx_core_module.html#worker_cpu_affinity
- https://nginx.org/en/docs/ngx_core_module.html#worker_rlimit_nofile
- https://nginx.org/en/docs/ngx_core_module.html#timer_resolution
- https://nginx.org/en/docs/ngx_core_module.html#worker_connections
- ....


**Ngx-php** https://github.com/rryqszq4/ngx-php#directives
php_ini_path ...;
init_worker_by_php ...;

I don't understand anything.

joanhey:
It's like node.js if they can add reuseport or cluster. As default don't work with this config.



Or like asp.net configuring the kestrel.

```
builder.WebHost.ConfigureKestrel(options =>
{
    options.Limits.Http2.MaxStreamsPerConnection = 256;
    options.Limits.Http2.InitialConnectionWindowSize = 2 * 1024 * 1024;
    options.Limits.Http2.InitialStreamWindowSize = 1024 * 1024;
```

MDA2AV:
worker_cpu_affinity auto
timer_resolution 1s
keepalive_requests 1000000
reuseport
backlog=65536
worker_connections 65536

these seem to be clearly tuned configs, they are documented in the docs but not typically used on a prod deploy, it should be nginx handling these not us setting deliberately.

nodejs doesnt work without reuseport, it is a core config, nginx does work fine without it and is the default config to not set it.

Those aspnet configs are to be removed from any prod type

joanhey:
**Nginx**
Nginx defaults are not specifically configured for a big server; they provide a generic, pre-baked configuration ready for most general use cases, including small to medium deployments.
- The default setup often includes conservative values like worker_processes 1 and worker_connections 1024
- **Production-grade traffic** typically requires tuning these parameters (e.g., setting worker_processes to the number of CPU cores) .
- While the architecture supports high concurrency, the default configuration itself is intended as a starting point rather than an optimized setup for large-scale servers. 

Later: 
- **worker_cpu_affinity auto now** it's not working, because the bench change to only use physical cores, because using also the logical cores, some _frameworks have worst_ results. A biased change, I don't know.
  Actually for this value we have this error at startup:
  ```
  sched_setaffinity() failed (22: Invalid argument)
  sched_setaffinity() failed (22: Invalid argument)
  sched_setaffinity() failed (22: Invalid argument)
  ...
  ```

- **timer_resolution 1s** have no advantage, we can remove it.
- **reuseport** all frameworks will be tuned.
- **worker_connections 65536** sorry, so we leave as 1024 x 1 worker_process, that never will be a production config. Also never will be used all this worker connections with this config.
- **keepalive_requests 1000000** that's a legacy option for old faulty frameworks, than fail after n requests. Nginx for me maintain too legacy options, this need to have a -1 value for a health framework.
   Actix and Kestrel have unlimited requests.
- **backlog=65536** this is fundamental for any language, platform,... because is used in the `listen()`. Without `listen()` we can connect to any socket :thinking: , and what is the synopsis for this call:
  `int listen(int sockfd, int backlog);`
  The rest of frameworks need to learn that.

**Nodejs**
By default, Node.js sockets do not have the reusePort option enabled (it defaults to false), meaning a new socket cannot bind to a port already in use by another process.

Etc, etc,..
I think that more frameworks need to learn from this `nginx.conf` file, also the AI (hello Benny).
But mark tuned this config, because the another frameworks don't use all the potential than they can. I don't think so.

joanhey:
About that config that is not "production".
What is this ? :thinking: 
```
worker_rlimit_nofile 65535;
worker_connections 65536;
...
```

Well, marked as "production":
**aspnet-minimal_nginx** 
https://github.com/MDA2AV/HttpArena/blob/main/frameworks/aspnet-minimal_nginx/proxy-production/nginx.conf#L16-L22

https://github.com/MDA2AV/HttpArena/blob/main/frameworks/aspnet-minimal_nginx/proxy/nginx.conf#L1-L6

....

joanhey:
What is production for me actually:

### Production
- H1 all with LTS
- All responses with gzip, br, ... if the client ask for them (99%).
- No H1 pipeline. No browser use pipeline, and only 2 esoteric proxy servers, than nobody use.
  The actual way is to use the H2 multiplex.
- Best configuration for the each server (manual or automatic). If the automatic configuration don't work correctly, we need to do it manually.
- ...

### Tuned
- Use caching
- Use static files from memory, bypass the IO to hard disk.
- Use simd json
- ...

This bench seems to biased to asp.net, if the framework don't use the same automatic config then is tuned.

> e.g. ASP.NET `MapStaticAssets`, `IMemoryCache`, automatic config,... the configs and rules than are based always 

We are developers, we create code, and the asp.net opinionated classes are not always the optimal or more performant.
We need to create our solutions, and try to be efficient with the code and than the rest learn.

If any dev want easy automatic solutions, they can check the code in this repo and choose.
Caddy server need less lines than nginx, because do it all automatic, but in big servers never is near to nginx in performance. 
Choose automatic or manual config (or use AI).

MDA2AV:
Yea these nginx flavour "frameworks" are going to be a pain in the ass, there are a trillion ways to configure nginx and every can claim they are production because there are use cases for all those configs. This is more of a nginx module than a framework which sort of invalidates prod/tuned logic applied to it because nginx is not a web framework. These prod rules are targeted to frameworks like Spring, Quarkus, Helidon, Elysia, ASP.NET, Laravel, FastHttp, Gin, etc which are a different category, for this reason we need a new category for anything that falls on reverse proxies, web servers like caddy, nginx, traefik, etc


So I'll just move whatever is nginx based to this new category, apples to apples

joanhey:
Like you want.
But a benchmark need to be fair and not biased.
It's the developer than need to decide which framework to use.

Thank you.

joanhey:
We can't move ngx-php to the gateway test, as minimum need 2 docker containers !!

MDA2AV:
So you think it is fair to have ngx-php and laravel under same category?

joanhey:
Any framework than use `fpm-php`  can never compare with any of the rest of frameworks in this bench.
But ngx-php, can't compare with pure Rust, C#, java, ... frameworks ??

MDA2AV:
Because it does not provide the same functionality. Oh cmon you know exactly what I am talking about, so anyone can just add a module to nginx or some low level engine to support basic features like routing and request parsing and compete with a full feature web framework?

joanhey:
Later I will add Symfony and Laravel to this bench, but not with `fpm-php` (we can do it for check the difference) but never will be performant with normal `fpm-php`.

Even Symfony and Laravel are always slow, is the same which platform they use.

joanhey:
If you have a Rust framework, you have all the possibilities  to configure in the best way to be faster than any module from nginx, than later create all the code in an interpreted language, not compiled !!

MDA2AV:
But you have to code in Rust to use it so it sucks already and nobody cares

Build a basic web api like we enforce here is doable fast in any language, if you have to build a large reliable and mantainable one nobody is going to use these Rust or shortcut modules

It's cool to see them go fast wow 3-4 million requests per second on baseline, cool for a benchmark but what I want to see is real web frameworks performance comparison like aspnet vs spring vs quarkus vs fasthttp vs laravel

joanhey:
> features like routing and request parsing and compete with a full feature web framework?

That is a framework, request and routing it's the core of any framework, and need to be fast. How you use it later is opinionated. The rest of classes on top are secondary.

> But you have to code in Rust to use it so it sucks already and nobody cares

Yes, but with ngx-php, we use plain and easy interpreted php, and we can use almost any framework on top.

joanhey:
I have ready 3 full PHP MVC frameworks with Adapteman, than use Workerman, to add to this bench.

But some devs, don't want to use these full beasts, they prefer use simpler solutions.

joanhey:
I said here what is production and tuned actually for me.
https://github.com/MDA2AV/HttpArena/pull/559#issuecomment-4276442567

But the language, platform, framework, ... it's a decision of the developer, not from us.

Perhaps "production" and "tuned" need to be changed for other more appropriate words.

MDA2AV:
The production criteria is simple: Default config.

If a framework needs to be tuned like nginx this isn't a web framework, it is a web server or something else, web frameworks adapt to the environment they run in and if we allow people to start configuring everything then whatever numbers we see on production are a byproduct of a config crafted to the benchmark

nginx and any of its modules cannot fit in the same category as a web framework and I won't mark as framework anything that is not battle tested with proper documentation like any vibe coded framework with 3 stars

joanhey:
Like you want.
But I don't think so.

joanhey:
Mark it as tuned !!
Now we can start to tune it.

Thank you.

MDA2AV:
Not as tuned, will be marked with new category Infrastructure

joanhey:
So OpenRestry will be in the same category. But not Rust frameworks. :thinking:

MDA2AV:
If workerman is allowed in production so is actix. 

nginx openresty, ngx_lua, h2o_mruby will fall together with nginx, h2o, traefik, caddy, pingora etc as they are extensions to it and configurations can be applied without having to be marked separately as tuned

MDA2AV:
and pipeline test will be removed in the future, no need to mention that all the time

joanhey:
:+1:

joanhey:
:+1: 

PD: https://github.com/MDA2AV/HttpArena/actions/runs/24627184967/job/72008157109

joanhey:
:sos:

joanhey:
:hand:

joanhey:
> As of mid-2026, Nginx is the leading web server, commanding approximately 31.5% to 35.1% of the market.

> Nginx: Dominates with 31.5%–41.7% depending on the survey (W3Techs vs. others), valued for high traffic efficiency and reverse proxy capabilities.

> In the niche of high traffic sites (first million), nginx dominates with a 55.96% share, far surpassing Apache.

joanhey:
Bye

PR #593, opened 2026-04-21
Add Lute runtime

Nicell:
## Description

Adds the [Lute](https://lute.luau.org/) runtime, built on top of the [Luau](https://luau.org/) language.

---

| Command | Description |
|---------|-------------|

---

<details>

```bash
```

</details>


MDA2AV:
Hey, nice PR but one detail, can you please not cache the JSON responses? Should compute and serialize it every request :D

Nicell:
@MDA2AV updated! do I re-run `/benchmark -f lute --save` now?

edit: oops didn't realize it just does is substring search lol

Nicell:
@MDA2AV if I switch from cached static files to reading from disk, can I switch this from tuned to prod considering the json cache is gone?

MDA2AV:
> @MDA2AV if I switch from cached static files to reading from disk, can I switch this from tuned to prod considering the json cache is gone?

Yes unless there is an official API for static file serving, in that case you need to use it. If there isn't then it is fine just serving from the disk, you can serve the pre compressed files and still be production too

MDA2AV:
Also if possible add yourself to "maintainers": [] in the meta.json so you're notified if anyone tries to modify this framework implementation

Nicell:
Updated @MDA2AV thanks. LMK if caching the headers is against the rules, seems like it isn't but please correct me.

MDA2AV:
> Updated @MDA2AV thanks. LMK if caching the headers is against the rules, seems like it isn't but please correct me.

Since it is marked as production this will take a bit longer to be approved, I'll need to check the framework repo and see if the practices are the prod intended. If there is no better option than to hardcode some headers beacuse the framework does not support the functionality yes it can be accepted because there is no better way

I should reply here max in 1-2 days whether we accept this as production, there are at least some base criteria, it must be a framework with proper request parsing and response building functionality plus having online documentation on best practices and general API.

Nicell:
If there's anything I can do to clear anything up let me know! I'm a maintainer of lute, which is very new, so understood if it doesn't fit the production label at the moment.

MDA2AV:
First of all, cool project. I see this is both a new language and a framework at the same time.

After reviewing there are a few points:

**Network model**
Lute's network layer is a Luau binding over uwebsockets (uws → usockets → libuv + boringssl + zlib) so h1 + websocket only, with fully-buffered request bodies, no h2 or h3, and no built-in compression or static-file middleware.                                                                                          
   
Concurrency is isolated vms per os thread (one Luau state per vm.create, fanned out with task.spawn and sharing the listen socket via reuseport=true), there's no shared memory across vms, so scaling to n cores means independent runtimes.

For outbound requests libcurl is used - good.

_Blocking rule violations:_
                                                                                        
  _1. Static handler - No inbuilt handler or API for this, so according to rules the only acceptable is to load files from disk every request, also no compression module so not allowed to use the pre compressed files.   
  2. Nightly build — Dockerfile pins 1.0.1-nightly.20260421 instead of stable 1.0.0._    
  
Gray area: explicit multi-vm fan-out + reuseport=true — Lute's only way to use all cores, but strict reading of baseline's "no worker count beyond defaults" may be violated, I'll say this is good for me.  - No clear reference on the docs whether reuseport is a default parameter we should always enable but seems like this is required due to the networking model.

My final take is that there is production potential, we can accept as production after fixing the 2 rule violations mentioned but my suggestion is to keep as it is as tuned and as the framework matures eventually we can move it to production.

@Kaliumhexacyanoferrat any take?

Nicell:
@MDA2AV There's a bug in 1.0.0 that blocked me using it on my local machine with this repo, so that's why it's on nightly. I can wait until 1.0.1 / 1.1.0 lands to update this PR if you'd like and then I can drop the static test (or maybe even upload? without streams that test feels a bit useless).

MDA2AV:
Ok, keep the nightly, adapt the static test and mark production

Nicell:
dropped static @MDA2AV! also /benchmark -f lute --save

thanks for the quick responses on this. cool project you have here!

MDA2AV (approved):

PR #599, opened 2026-04-22
Improve WTX

c410-f3r:
## Description

Looks like the performance of `WTX` is terrible, which is strange given other similar benchmarks and the time I spent optimizing it.

Well, this is the last attempt without diving into profiling applications . Let's see if things will improve.


MDA2AV:
I'll add some more frameworks on this test for a broader comparison, could be a load generator issue or not optimal config on connections/streams

c410-f3r:
Hum, seems unlikely given the results of the other HTTP/2 frameworks.

It is probably a bad default HTTP/2 configuration like the maximum number of concurrent streams or a recent change I made in the `WTX` project.

c410-f3r:
Hum, let me tweak some HTTP/2 parameters to see their runtime impact. I will be back after a couple of hours.

MDA2AV:
Added a few more h2c entries and more conn variety for better comparison

https://www.http-arena.com/leaderboard/#v=h2&t=baseline-h2c&type=infrastructure,production,tuned

c410-f3r:
> Added a few more h2c entries and more conn variety for better comparison
> 
> https://www.http-arena.com/leaderboard/#v=h2&t=baseline-h2c&type=infrastructure,production,tuned

Capture d’écran du 2026-04-22 15-58-59

Wow... The performance of h2o is really impressive

Capture d’écran du 2026-04-22 16-00-32

Not so much in the JSON scenario, strange.

c410-f3r:
Nah, I give up. That was the best I could do given the constraints.

At least it is clear that HTTP/2 parameters weren't the problem.

c410-f3r:
Perhaps I will get back after a couple of months. The code changes are still worth merging though.

MDA2AV:
> Not so much in the JSON scenario, strange.

 Yea, the json implementation is really bad for it, jsut threw it in

MDA2AV (approved):

PR #605, opened 2026-04-22
[python] FastPySGI: update version to v0.6 and enable JSON-TLS

remittor:

| Command | Description |
|---------|-------------|

### DIFF
--- a/frameworks/fastpysgi-asgi/Dockerfile
+++ b/frameworks/fastpysgi-asgi/Dockerfile
@@ -3,5 +3,5 @@ WORKDIR /app
 COPY requirements.txt .
 RUN pip install --no-cache-dir -r requirements.txt
 COPY . .
-EXPOSE 8080 8081
+EXPOSE 8080
 CMD [ "python", "app.py" ]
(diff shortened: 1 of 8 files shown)


MDA2AV:
I think the verbose does not yet work on all tests, so you probably hit them.

I'll look it up later today

remittor:
> I think the verbose does not yet work on all tests, so you probably hit them.

It looks like my latest changes introduced a bug into the ASGI mode. I don't know how to reproduce it yet. I'll have to think about it...

remittor:
@MDA2AV , I still haven't been able to reproduce the issue locally.
I just downgraded the Python version in Docker to 3.9 (that's what I'm testing on), and all the checks passed right away.
So you can commit this PR, and I'll try to fix the issue in v0.7 on other Python versions.

remittor:
It's clear to see how much slower Python 3.9 is than 3.13

remittor:
I reproduced this bug. It only occurs after test `pipeline` ;-)
Now I'll be looking for the cause of the bug locally. And this bug doesn't exist on Python 3.9

MDA2AV (approved):

PR #606, opened 2026-04-22
SlimeWeb Python -few improvement changes

ATOMMAX-2001:
+ precompress lookup in static 
+ pong frame for ws
+ async removed lock logic
+ reduce no of worker for both async and sync thread to prevent too much context switching.

Hoping to see few more RPS than previous run.


ATOMMAX-2001:
well 10k rps is increased  interesting..

```plain
[info] waiting for server...
[warn] slimeweb did not come up for echo-ws; skipping
```

Hey @MDA2AV , can you tell me how you’re checking whether the echo-ws server is up or not? In all three cases, it’s showing the same log. I thought there was an issue with the ping-pong frames, so I fixed that, but the problem still persists. Since it passed validation phase, I’m a bit confused.

MDA2AV:
The readiness check is just curl -s -o /dev/null --max-time 2 http://localhost:8080/ws looped for 30s. It accepts any HTTP response code. It sounds like your /ws handler closes the socket on a plain GET (no Upgrade: websocket header) instead of returning an HTTP error. Have it return a 400 or 426 on non-upgrade requests and the probe will pass.

ATOMMAX-2001:
Ahhhhh, Now I understood. Thanks

MDA2AV (approved):

PR #613, opened 2026-04-23
[roda] Mark as production

p8:
Roda uses standard framework configuration.


Kaliumhexacyanoferrat:
The compressed JSON test is implemented in the handler:

```ruby
      if accept_encodings = r.headers['Accept-Encoding']
        type = accept_encodings.split(',').map(&:strip)
        if type.include? 'gzip'
          sio = StringIO.new
          gz = Zlib::GzipWriter.new(sio, 1)
          gz.write(result)
          gz.close
          response[RodaResponseHeaders::CONTENT_ENCODING] = 'gzip'
          result = sio.string
        end
      end
```

The rules state

> Must use the framework standard JSON serialization and the framework or engine's **built-in response compression (middleware, filter, or equivalent).** No pre-compressed caches, **no bypassing the response pipeline**.

There are `:gzip` and `:br` plugins which should be able to dynamically compress responses, as far as I understand the documentation.

Kaliumhexacyanoferrat (approved):

PR #614, opened 2026-04-23
[sinatra] Mark as production

p8:
Sinatra uses standard framework configuration.


Kaliumhexacyanoferrat:
Just a note, you might want to add `--deployment` to the `bundle` call in docker.

p8:
> Just a note, you might want to add `--deployment` to the `bundle` call in docker.

Thanks. I've added the deployment config option (--deployment as a flag is deprecated).

MDA2AV (approved):

PR #619, opened 2026-04-23
[rails] Let thruster handle gzip

p8:


p8:
This fails because compression is only turned on if the response size is greater than 1Kb.
https://github.com/klauspost/compress/blob/v1.17.4/gzhttp/compress.go#L50-L54
The idea is that compression doesn't make sense if the response is smaller than the maximum package size of 1500 bytes.

MDA2AV:
Hmm makes sense, I guess for this test we should use larger counts only

p8:
Closing, as thruster has been removed for now.

MDA2AV:
btw, json-comp test was changed to have larger sizes only, this should have fixed the issue with not compressing because obejct too small

p8:
@MDA2AV Ah ok, good to know! Thanks!

PR #641, opened 2026-04-26
refactor: improve local development resilience and add SKIP_TUNE

kevmoo:
- Make CPU pinning (cpuset-cpus) dynamic based on host core count
- Add SKIP_TUNE environment variable to bypass sudo-required system tuning
- Update documentation to reflect new dynamic defaults and configuration options

### DIFF
--- a/scripts/benchmark.sh
+++ b/scripts/benchmark.sh
@@ -90,7 +90,11 @@ cleanup_all() {
     docker volume prune -f >/dev/null 2>&1 || true
     docker image prune  -f >/dev/null 2>&1 || true
 }
-trap 'cleanup_all; system_restore' EXIT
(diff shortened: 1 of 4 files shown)


MDA2AV:
Typically for local development you'd want to use [the lite script version benchmark-lite.sh](https://www.http-arena.com/docs/running-locally/scripts/benchmark-lite/), it doesn't pin to specific cores and uses a much lower connection count. The system_tune is still called though, that can be adjusted.

Wouldn't it be better to extend the lite version or do you see a need to be able to run the original benchmark script locally too?

kevmoo:
It's nice to be able to debug high connection counts (like for race conditions) without crashing on a 16-core laptop.

The main entry point should "just work" for everyone, I think!

MDA2AV (approved):

PR #666, opened 2026-04-27
[roda] Update puma to 8.0

p8:


p8:
Argh, that contained a commit I did't want yet.

p8:
Not sure about the regressions in the static calls, but it is what is is.

MDA2AV:
There is no regression in static, seems that all frameworks are having regression on this test. Kernel version was updated, need to investigate this.

Edit: Not all frameworks, just some which is interesting

p8:
Ah, that is interesting.

MDA2AV (approved):

PR #672, opened 2026-04-29
[roda] Use Puma's "mark_as_io_bound" option

p8:
Use Puma 8's new 'mark_as_io_bound' option allow more threads on the IO heavy upload call.
https://github.com/puma/puma/pull/3816


p8:
Enabled it for all calls to see the impact.

p8:
I've only turned it on for "/baseline" as that seems to have a real improvement.

MDA2AV (approved):

PR #685, opened 2026-05-03
[rails] Upgrade Puma to 8

p8:
Use Puma 8's new 'mark_as_io_bound' option to allow more threads on the IO heavy calls.
It seems only the baseline call benefits from this.


p8:
Tried enabling mark_as_io_bound for all tests.

MDA2AV (approved):

PR #698, opened 2026-05-11
[roda] Call public/static route as last

p8:
The static route checks if files exist, which can be expensive. Instead, match this route only if others don't match.


MDA2AV:
> /benchmark -f roda —save

use --save, the em dash doesnt work btw

p8:
> > /benchmark -f roda —save
> 
> use --save, the em dash doesnt work btw

Ah, that was an autocomplete on the iPhone
/benchmark -f roda --save

MDA2AV (approved):

PR #712, opened 2026-05-16
Add rack-iodine engine test

p8:


p8:
With threads set to 3 instead of 8:
/benchmark -f rack-iodine

p8:
With threads set to 4 instead of 3:
/benchmark -f rack-iodine

MDA2AV (approved):

PR #739, opened 2026-05-19
Add zeemo-tuned — first Zig tuned-category entry

skylightis666:
## Description

First Zig entry in the **`tuned`** category. Built on the [`zeemo`](https://github.com/skylightis666/zeemo) Zig HTTP library — same author and same core io_uring transport as the existing `zeemo` (engine) entry, but reworked as a proper framework:

- Public Zig API (`Server.init/.get/.post/.run`, `Request`, `Response`, comptime JSON serializer, static file mount)
- JSON responses are serialized per request from native Zig structs via the library's comptime-reflection serializer. No pre-rendered fragments, no per-(count,m) response caching — strictly the runtime work that the `json` profile is designed to measure.
- Static files are loaded into memory at startup and the full HTTP/1.1 response is pre-baked per (file, encoding) variant — allowed for `tuned` per the [static profile rules](https://www.http-arena.com/docs/test-profiles/h1/isolated/static/implementation/) ("May cache files in memory at startup, use memory-mapped files, pre-rendered response headers, or any caching strategy. May serve pre-compressed files (.gz, .br) from disk via any mechanism.").
- The framework dependency is a real, published library: https://github.com/skylightis666/zeemo (pinned in `build.zig.zon` via `zig fetch --save`).

## Subscribed profiles

- `baseline` — `GET/POST /baseline11?a=…&b=…`
- `pipelined` — `GET /pipeline`
- `limited-conn` — same `/baseline11` endpoint, connection-churn pattern
- `json` — `GET /json/{count}?m=…` with per-request struct walk
- `static` — `GET /static/*` with `.br`/`.gz` variant selection by `Accept-Encoding`

## Local validation

Full `scripts/validate.sh zeemo-tuned` suite passes — **22 checks, 0 failures** (baseline GET/POST/chunked + Content-Type, anti-cheat random inputs, 4 TCP-fragmentation cases, pipelined + Content-Type, JSON shape for 4 counts × multipliers + Content-Type, static Content-Type for css/js/json, 20 file sizes uncompressed, 15 compression variants, 404 for missing).

## How tuned differs from the existing engine entry

- Goes through a framework HTTP server (`zeemo.Server`), not raw socket dispatch
- JSON serialized per request from live struct data (no pre-rendered prefixes)
- Same io_uring transport, multishot accept, fork-workers with `SO_REUSEPORT`, parser

PR Commands — comment to trigger (requires collaborator approval):

| Command | Description |
|---------|-------------|
| `/benchmark -f zeemo-tuned` | Preview run |
| `/benchmark -f zeemo-tuned --save` | Run and save results |


MDA2AV:
Tuned or Production entries must be actual frameworks, this is still very early work, it needs more than just routing handlers to the engine. At least having some proper documentation on how to use it to build an API. Rule of thumb is "would I use this to build something I can deploy to prod and be confident about it?"

skylightis666:
Fully fair — agree it's early for "framework" status. Converting back to draft and keeping it open as the tracking PR while I bring zeemo up to that bar. Plan:

**Framework features**
- Middleware pipeline (`server.use(...)`, ordered, with `req.next()`-style chaining)
- Error handling — `server.onError(fn)`, structured `Response.problemJson(rfc7807)`, default 4xx/5xx pages
- Body parsing helpers — JSON-in, form, multipart
- Graceful shutdown on SIGTERM
- Logging interface — pluggable sink, request-scoped fields
- Request context / locals — handler-local typed state

**Additional profile coverage**
- `upload` — streaming body API (currently capped at small bodies)
- `json-comp` — gzip + brotli via `@cImport`, honoring `Accept-Encoding`
- `json-tls` — TLS server through BoringSSL bindings
- `crud` — Postgres client + cache-aside read pattern (covers the most "real-world" workload)

**Documentation**
- Getting started tutorial (write your first API end-to-end)
- API reference for every public symbol
- Architecture overview (io_uring loop, fork-workers, lifecycle)
- Production deployment guide (Dockerfile, systemd, tuning knobs and their tradeoffs)
- Cookbook (auth, validation, file upload, REST CRUD patterns)

**Real example**
- Partial FHIR (HL7) Patient resource API — search by name/birthdate, create/update/delete, content negotiation, structured error responses. FHIR is what you'd ship in a healthcare backend, so it forces the framework through realistic JSON-in, validation, search params, paginated responses, and proper error codes.

Keeping PR #739 in draft until all of those land so we don't lose the benchmark-runner association. Will ping when ready for a fresh look.

MDA2AV:
Closing for now, reopen when ready

PR #790, opened 2026-06-03
ktor results

MDA2AV:
@bjhham fyi 

updating ktor results on leaderboard


bjhham:
Could you fire off another one of these to include the updates? https://github.com/MDA2AV/HttpArena/pull/794

MDA2AV:
Started now, results should be available in 20 min

PR #807, opened 2026-06-04
Add pico (V) framework

enghitalo:
## Add `pico` (V)

A raw [V](https://vlang.io) server on the `picoev` non-blocking event loop + `picohttpparser` (both stdlib). One process per core shares the listen socket via `SO_REUSEPORT`.

### Profiles implemented

| Profile | Endpoint |
|---|---|
| `baseline` | `GET/POST /baseline11` (decodes chunked) |
| `pipelined` | `GET /pipeline` |
| `json` | `GET /json/{count}?m=M` over `/data/dataset.json` |

### Notes

- Pinned **V 0.5.1** (prebuilt release binary).
- JSON serialized manually (precomputed per-item prefixes + `strings.Builder`), no per-request reflection.
- **DB profiles are not subscribed**: picoev is a single-threaded event loop and the stdlib `db.pg` driver is blocking, so a query would stall the loop. A non-blocking PG path would be needed.

### Validation

Verified against the HttpArena image + `data/dataset.json`: baseline (GET/POST, chunked, randomized anti-cheat, `text/plain`), pipelined, and json (correct across counts/multipliers, `application/json`) all pass.

(`./scripts/validate.sh` couldn't run end-to-end locally because host port 5432 was held by an unrelated container; checks were run against the built image.)

### DIFF
--- a/frameworks/pico/.gitignore
+++ b/frameworks/pico/.gitignore
@@ -1,0 +1,2 @@
+server
+*.so
--- a/frameworks/pico/Dockerfile
+++ b/frameworks/pico/Dockerfile
@@ -1,0 +1,25 @@
+FROM debian:stable-slim AS build
+
+RUN apt-get -qq update && \
+    apt-get -qy install --no-install-recommends \
+        ca-certificates curl unzip build-essential && \
+    rm -rf /var/lib/apt/lists/*
+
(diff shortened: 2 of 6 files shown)


MDA2AV:
Hi, any entry that can't support baseline tests needs to be marked as engine (type) in the meta.json

PR #809, opened 2026-06-05
[Rust] Add WTX - WebSocket

c410-f3r:
## Description

Adds a new WebSocket implementation based on `WTX`


c410-f3r:
Should be good now 

/benchmark -f wtx-ws --save

c410-f3r:
```
echo-ws-pipeline 	16384 	0 	0.0% 	0MiB 	NEW 	NEW
```

Erhh... Did a last minute modification. Let's see if this case will be resolved.

/benchmark -f wtx-ws --save

c410-f3r:
👍

MDA2AV:
This should be marked as type: engine
rest seems good to go

c410-f3r:
Oops... I should have used the higher level abstractions instead of manually accepting connections.

Let me know if the code is now in good shape for the current label.

MDA2AV:
Both production and tuned are framework territory/tier - requires a framework«s high level APIs (routing, middleware, handlers, etc) which is ofc always a gray territory as different technologies have their unique way of implementing web frameworks. 

WTX is as described as collection of low-level transport implementation tools. This entry reflects that - drives a bare frame echo with no router or middleware which leans to the engine definition, we are essentially measuring tokio performance here

c410-f3r:
I see. The production tier requires a router/middleware.

Next time such a feature will be implemented. Changing to `engine` for now.

MDA2AV:
> I see. The production tier requires a router/middleware.
> 
> Next time such a feature will be implemented. Changing to `engine` for now.

It's not just that, production is reserved for frameworks with a strong ecossystem, currently we have some "production" entries that aren't really fit to be there and will be adjusted before the first official round, probably we will also change this classification system and the production equivalent will be only for frameworks with a community, ecossystem, full feature and an active team working on it, we will still be able to compare these with other frameworks on leaderboard but the distinction must exist because a production framework has a lot more constraints and implementation details than others

c410-f3r:
OK. All projects associated with WTX were converted to `engine`.

MDA2AV (approved):

PR #825, opened 2026-06-06
[Experiment] Modify linger parameter in HTTP/2 (WTX)

c410-f3r:
## Description

Requesting a run to see the impact of this change.


c410-f3r:
Thanks

PR #843, opened 2026-06-08
Make composite table horizontally scrollable everywhere

MDA2AV:
The desktop fix set the table wrapper to overflow-x:hidden and scrolled it by setting wrap.scrollLeft from the synced bar. Programmatic scrollLeft on an overflow:hidden box is browser-flaky - some browsers move the bar but never scroll the table content (reported on the live site).

Switch the wrapper to overflow-x:auto so it is a real scroll container: the synced bar, mouse wheel, trackpad and finger-swipe all scroll it reliably. Hide the wrapper's own horizontal scrollbar (webkit height:0; thin elsewhere) so only the synced top/bottom bar is shown, and keep the vertical scrollbar. Replaces the earlier touch-only media query.

### DIFF
--- a/site/static/new-leaderboard/index.html
+++ b/site/static/new-leaderboard/index.html
@@ -285,7 +285,12 @@
 
 
   /* ── Composite matrix ────────────────────────────────────── */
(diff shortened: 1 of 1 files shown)


MDA2AV:
Duplicate of #842 (same fix, but #842 is branched off current main and has a clean diff). Closing this one.

PR #845, opened 2026-06-08
Fix table hscroll

MDA2AV:


MDA2AV:
Duplicate of #844 (conflicting because this branch's base was squash-merged via #842). The same mobile full-height change is in #844, which is clean off current main. Closing.

PR #859, opened 2026-06-11
Add swerver-tuned (emerging, mode: tuned)

justinGrosvenor:
Adds **swerver-tuned**, a tuned-mode entry for the [swerver](https://github.com/justinGrosvenor/swerver) engine (Zig HTTP/1.1+2+3 server). Mirrors the existing `swerver` engine entry but in `mode: tuned`:

- **Pre-encoded response registry OFF** (`preencoded: false`) — no pre-computed response bodies, per tuned rules. Every route goes through the normal pipeline.
- **Precompressed static serving** — serves `.br`/`.gz` siblings when `Accept-Encoding` allows (Content-Encoding + Vary), falling back to identity.
- **In-memory static cache** (`cache_static_files: true`) — files cached per-worker on first serve.
- **Native async PostgreSQL client** (wire v3, SCRAM-SHA-256, TLS, pipelining + prepared-statement cache, park/resume so queries never block the reactor) backing two PG profiles:
  - `async-db` — `SELECT ... FROM items WHERE price BETWEEN $1 AND $2 LIMIT $3`, `rating_score`/`rating_count` nested into a `rating` object.
  - `fortunes` — all rows + a runtime-injected row, sorted, HTML-escaped.

`frameworks/swerver-tuned/` pins swerver `v0.1.0-alpha.17` (immutable tag). The engine entry is unchanged.

### Validation
`scripts/validate.sh swerver-tuned` → **57 passed, 0 failed** across all 16 subscribed profiles (baseline, pipelined, limited-conn, json, json-tls, upload, static, baseline-h2, static-h2, baseline-h2c, json-h2c, json-comp, baseline-h3, static-h3, async-db, fortunes).

Happy to adjust naming/config to match conventions.


MDA2AV:
Ok, so this is a submission to framework grounds so there are some extra requirements that engines don't have. I took a look at Swerver repo and it is still not framework level, some obvious gaps like no proper documentation on how to use it and very narrow test coverage makes this project still not very usable by someone who wants to build something. There are other factors but these two (specifically no documentation) are heavy blocker for a framework entry, I can't evaluate something if there is no documentation on how to use it. It is expected that framework (flagship or emerging) entries can cover most of a user needs in terms of functionality/middleware to build a reliable deployable service.

Regardless of that a new type will soon be added to frameworks tier, it will be along the lines of "experimental" where frameworks that are still very early or have no community/user base can fit in. I think Swerve could be a good contender for it with a little bit more maturity.

I understand this is a bit of a bummer and I hate doing gatekeeping but it's all about having a fair comparison in the long run. Composite scores are relative based and if we add something in the lines of an engine performance, that will directly affect every other score, when we build a project with a benchmark test already in mind that is likely to overfit our solution to that specific scenario which is fine for engine entries but not frameworks. 

One example is the json test, it is a responder shaped to this benchmark's exact inputs and validator
`std.fmt.bufPrint(buf[off..], "{{\"id\":{d},\"name\":\"{s}\",\"category\":\"{s}\",...,\"total\":{d}}}", ...)
`

justinGrosvenor:
Thanks for the feedback. Not really a bummer per se, but it would be nice to be ranked for features that go beyond an engine,  so the experimental category sounds perfect. Still lots of work to do to get swerver to it's final form. The long goal is a single binary edge runtime covering what you'd otherwise deploy nginx, APISIX, oathkeeper, Varnish, and PgBouncer for, plus a few things none of them do,  like x402 payments.

Only thing I would point out is test coverage: it's broader than it looks from the top level . ~630 tests run under a five configuration build-flag matrix in CI (≈3,100 test executions per push), but I take the underlying  point, what's missing is user-facing validation: examples and integration tests that double as usage documentation. All of the other points are right on and I appreciate you pointing them out. They are helpful for me moving the project forward.

looking forward to the experimental category. will resubmit when that goes live.

Cheers

PR #866, opened 2026-06-13
vanilla: split into vanilla-epoll and vanilla-io_uring

enghitalo:
## What

Replaces the single `vanilla` framework entry with two engine variants that exercise the server's two Linux I/O backends:

| Entry | `engine` | `main.v` backend |
|---|---|---|
| `vanilla-epoll` | `epoll` | `io_multiplexing: .epoll` |
| `vanilla-io_uring` | `io_uring` | `io_multiplexing: .io_uring` |

## Why

[vanilla](https://github.com/enghitalo/vanilla) ships both an epoll and an io_uring backend, selected at runtime via one `ServerConfig` line. Splitting the entry lets the leaderboard compare the two backends head-to-head under identical handlers.

## Details

- Both variants share identical handlers, profiles, `Dockerfile`, and the pinned `enghitalo/vanilla` module — the **only** functional difference is the backend line in `main.v`.
- Setting `engine: io_uring` in `meta.json` makes the harness run that variant with `--security-opt seccomp=unconfined` and `--ulimit memlock=-1:-1` automatically ([`scripts/run.sh`](https://github.com/MDA2AV/HttpArena/blob/main/scripts/run.sh)).
- `site/data/frameworks.json` regenerated via `scripts/rebuild_site_data.py` (frameworks-only) — clean swap, no other entries touched.
- Historical result data under `site/data/*-NNN.json` left untouched (immutable run snapshots).


MDA2AV:
Only one --save per PR, everytime it runs, it clears previous --save commited data to avoid stale numbers from previous runs on the PR

MDA2AV (approved):

PR #877, opened 2026-06-15
vanilla: json-comp cache, zero-alloc routing, DB prepared statement + HTML escape

enghitalo:
## Summary

Allocation/CPU-reduction **and large-I/O** follow-ups to the zero-alloc handler (#866), in `frameworks/vanilla-epoll/main.v` (+ `vanilla-io_uring` where noted) — consolidated into one PR.

### Allocation / GC-pressure (both backends)

1. **json-comp gzip cache** — the response for a given `(count, m)` is deterministic and gzip CPU dominates, so the complete gzipped response is cached per `(count, m)` (bounded map + RwMutex) and appended on a hit. Compress once, reuse.
2. **Zero-alloc routing** — route on the path before `?` via a `tos()` view into the request buffer instead of `all_before('?')`'s per-request copy.
3. **Precomputed query keys** — `qint`/`qstr` take a precomputed `const []u8` (`qk_*`) instead of `key.bytes()` per call; path ints (`/json/<n>`, `/crud/items/<id>`) parsed in place (`parse_u_at`) instead of `route[n..]` substrings.
4. **DB: prepared statement** — async-db uses `PQprepare`/`PQexecPrepared` (lazily prepared per pooled connection) instead of `exec_param_many`'s per-request server-side SQL re-parse.
5. **DB: single-pass HTML escape** — `escape_html` (fortunes) does one pass with a no-alloc fast path instead of `replace_each`'s five full-string passes.
6. **Hot-path byte-write armor** — replace single-element `out << b` in the JSON paths with `push_many`/indexed writes (`wi` itoa into a stack scratch + one `push_many`; fused object `},` separators). Single-element `<<` is **4–7× slower** on post-0.5.1 V ([vlang/v#27468](https://github.com/vlang/v/issues/27468)); this keeps the hot path fast across V versions. Output is byte-identical (verified across counts 0..4096 + edge ints).

### Large I/O (epoll only; io_uring is a follow-up)

7. **`/upload` answered by Content-Length (streaming drain)** — the engine now drains large request bodies into a fixed buffer (recv + discard) instead of buffering the whole 20 MB into a per-conn read buffer (which grew into a big *scanned* GC block → 64-core stop-the-world cliff). The handler answers from `req.content_length()` (lib change merged as enghitalo/vanilla#31), falling back to `body.len` for small/buffered bodies (crud/json POSTs unaffected — only >1 MiB drains). **Local (vs the upload leader zix):** single-conn 45 req/s / 895 MB/s (zix 45 / 903); 32-parallel **303 req/s / 6.1 GB/s** (zix 336 / 6.7) — within ~10%, the 86× gap is closed; server RSS **12 MB** (was ~928 MB while buffering).
8. **Static assets via `sendfile(2)`** — preload each asset's `O_RDONLY` fd + a precomputed response head; `out << f.header` then `core.queue_file(f.fd, 0, f.size)` streams the body zero-copy from the page cache, so the write buffer never grows into a large scanned block (the static "cliff"). **Local vendor.js (307 KB, 64 conns): 25.7K → 59.3K req/s (2.3×)**, 7.36 → 16.97 GB/s; md5-identical, keep-alive OK.

### Build

9. **Pin V 0.5.1, built from source** (pinned-vc bootstrap `vlang/vc@f461dfeb`) — reproducible builds, and avoids the post-0.5.1 master `<<` regression (#27468) that otherwise slows every path. Matches the-benchmarker pin (the-benchmarker/web-frameworks#9466).

## Arena results (vanilla-epoll, 64-core, Δ vs #866's zero-alloc baseline)

| Test | Conn | RPS | Δ RPS | Δ Mem |
|---|---|---|---|---|
| **upload** | 32 | **11,109** | **+36,930%** | −89.5% |
| **upload** | 256 | **12,660** | **+22,507%** | −85.4% |
| **static** | 6800 | **1,168,588** | **+460.2%** | −90.1% |
| **static** | 4096 | 1,142,779 | +329.4% | −89.9% |
| **static** | 1024 | 1,075,725 | +238.1% | −87.1% |
| **pipelined** | 512 | **35,233,232** | **+1,380.6%** | −9.4% |
| **pipelined** | 4096 | 34,680,659 | +1,406.7% | −3.2% |
| **json-comp** | 4096 | **844,718** | **+1,380.1%** | −0.5% |
| **json-comp** | 16384 | 878,355 | +1,142.3% | −7.5% |
| **json-comp** | 512 | 496,559 | +914.7% | −33.3% |
| **json** | 4096 | 891,795 | +84.0% | +2.3% |
| **baseline** | 512 | 650,591 | +39.7% | −4.2% |
| **baseline** | 4096 | 621,002 | +44.7% | −2.6% |
| **crud** | 4096 | 155,943 | +30.4% | +8.1% |
| **fortunes** | 1024 | 3,894 | +46.7% | −57.6% |
| limited-conn | 4096 | 370,133 | +22.0% | −2.2% |
| api-16 | 1024 | 21,254 | +7.6% | −1.1% |
| async-db | 1024 | 11,094 | +2.1% | −4.0% |
| api-4 | 256 | 33,407 | +1.7% | −4.2% |

Highlights: **upload now leads the profile** — the streaming drain (#7) took it from ~30 to **11,109 req/s**, past the prior leader (~4.3K), at **−89% memory**. **static +460%** at **−90% memory** (sendfile(2), #8 — the buffer-growth cliff is gone). **json-comp leads** (+1,380%, gzip-response cache). pipelined +1,407% (zero-alloc routing + `<<`-armor). The DB-bound profiles (api-4/16, async-db) are flat by design — #877 doesn't touch the DB engine; closing those is the async DB driver's job (below).

## Lib dependencies (all merged to enghitalo/vanilla `main`)

- **#29** — guards the `.noscan_data` flag behind `$if vanilla_noscan ?` so the lib builds on 0.5.1.
- **#31** — streaming drain for large bodies (`frame_head_len` + `HttpRequest.content_length()`), which #7 here depends on.

The DB-bound profiles stay flat because they're driver-bound (`db.pg`'s synchronous text protocol; `veb` on the same stack is slower, so vanilla already maxes it). Closing that gap is the **async, epoll-integrated DB driver**: the async runtime (`ac.watch(fd, …)` + continuations) is now merged to enghitalo/vanilla `main` (#34 → #36 → #37, Linux epoll + macOS kqueue), and a measured PoC showed ~6× per-thread on the DB path. The remaining piece — a thin `db_async` consumer on that runtime — is tracked in enghitalo/vanilla#32. Correctness verified for all paths here.

Supersedes #878 (folded in here).


enghitalo:
Closing as **superseded by #888** (merged). #888 took a different, faster DB architecture — a native `pg_async` wire client with cross-request pipelining (async-db ≈269k RPS) — whereas this branch's stdlib `db.pg` prepared-statements path benched ≈10k RPS (-3.9%). The non-DB wins here (json-comp gz cache, single-pass HTML escape, zero-alloc routing) already exist in main via #888's `WorkerCtx`/zero-alloc design (`escape_html_into`, the shared `gz` cache, `qstr_slice`/`parse_i64_slice`). Remaining crud/fortunes render allocations are handled in #894. Nothing here is lost.

PR #878, opened 2026-06-15
vanilla: DB path — prepared statement (async-db) + single-pass HTML escape

enghitalo:
## Summary

Two DB-path improvements in `frameworks/vanilla-{epoll,io_uring}/main.v`:

1. **Prepared statement for async-db.** `exec_param_many` (`PQexecParams`) re-parses the SQL **server-side on every request**. Switched to a PostgreSQL prepared statement (`PQprepare` + `PQexecPrepared` via `db.pg`), **lazily prepared on each pooled connection's first use** — prepared statements are per-session and the pool reuses sessions, so after warmup every connection serves `exec_prepared`.
2. **Single-pass HTML escape (fortunes).** `escape_html` did five full-string `replace_each` passes; now it's one pass with a fast path that returns the message untouched (no allocation) when there's nothing to escape.

## Measured (local, 16-core loopback, gcannon, single listener)

| profile | before | after | Δ |
|---|---|---|---|
| **fortunes** | ~5.6K | ~7.1K | **+27%** (escape rewrite) |
| **async-db** | ~18.0K | ~19.6K | **+9%** (prepared statement) |

## Honest scope

These are **incremental**. The DB profiles remain bound by the stdlib `db.pg` driver (text protocol, lazy pool) — confirmed by `veb` (same V + `db.pg`) being *slower* than vanilla on async-db. The top DB engines (swerver, genhttp) use native binary-protocol drivers with persistent prepared statements + pipelining; closing that gap needs a faster pg driver (a separate, larger project), not framework tweaks. Correctness verified (async-db rows + count; fortunes renders the same rows). Both epoll and io_uring variants.

### DIFF
--- a/frameworks/vanilla-epoll/main.v
+++ b/frameworks/vanilla-epoll/main.v
@@ -386,7 +386,31 @@ fn (mut sh Shared) fortunes() string {
 }
 
 fn escape_html(s string) string {
-	return s.replace_each(['&', '&amp;', '<', '&lt;', '>', '&gt;', '"', '&quot;', "'", '&apos;'])
+	// Fast path: most fortune messages contain no special characters, so return
+	// the original with no allocation instead of replace_each's 5 full-string
+	// passes (each scanning + reallocating). Only escape when there's something to.
+	mut needs := false
+	for c in s {
+		if c == `&` || c == `<` || c == `>` || c == `"` || c == `'` {
+			needs = true
+			break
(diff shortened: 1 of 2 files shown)


enghitalo:
Folded into #877 so everything lands in a single PR (json-comp cache + zero-alloc routing/keys + DB prepared statement + single-pass HTML escape).

PR #884, opened 2026-06-17
vanilla-epoll: re-enable async-db (pg_async park/resume) on the flat-state runtime

enghitalo:
**Experiment — benchmark vs #877 (the sync baseline).**

## What
Switches the `vanilla-epoll` DB path from the synchronous `db.pg` connection pool back to the **native `pg_async` park-and-resume** model: per-worker `PgPool` via `make_state`, an `async_handler` that submits the query and `ac.watch(pg_fd)` → `.suspend`, and a single `on_db_ready` continuation that pumps the result, renders by kind, and releases the connection. Non-DB routes (`/pipeline`, `/baseline11`, `/json`, `/static`, `/upload`) return `.done` immediately.

## Why
On the sync path, `fortunes` (~3.0k rps) and `async-db` (~10.9k rps) are capped at **~16-way concurrency** by thread-per-core blocking — each worker thread blocks on `acquire → exec → release`, so the 64-connection pool sits ~75% idle and the server runs at **~460% CPU** (≈11 cores idle, parked in `libpq recv` waiting on Postgres). Park-and-resume frees the worker to keep many queries in flight → uses the whole pool → the model the current `async-db` leader (swerver, ~370k) uses.

## Why now (it was reverted before)
This is the #32 conversion that was reverted (regressed the arena) because the **old map-based reactor + a per-request malloc** per watch cost more than the sync path won at sub-ms local-PG latency. enghitalo/vanilla#41 replaced that with a **flat fd-indexed reactor** (no hashmap, no per-request allocation) — the exact overhead that sank it. This PR is the experiment to confirm #41 turns the regression into a win.

## Notes for review
- **Stacked on #877**: this branch is #877 (json-comp gzip cache, route-slice, `/pipeline` short-circuit) **plus** the async DB re-conversion — the delta vs #877 is the DB path going async. Build verified clean against the flat-state runtime.
- The async framework code itself is the previously-validated #32 conversion (it passed `validate.sh` before); only the runtime underneath changed.

## What to watch in the benchmark
- **Should rise:** `fortunes`, `async-db`, `crud` (target 4–12× as concurrency lifts from ~16 to pool size).
- **Must NOT regress:** `baseline`, `json`, `json-comp` (#2), `pipelined`, `static` (now ~1.2× from #1) — these are non-DB and return `.done` immediately, so they should match #877.

`/benchmark -f vanilla-epoll`


enghitalo:
Closing — already in main. #888 ("cross-request pipelining for async-db, stacked on #884") was built on this branch and merged (squash), so this PR's content — re-enabling async-db via `pg_async` park/resume on the flat-state runtime — landed in main under a new hash. `main` now has the full design: `start_async_db`, `park`/`watch_persistent`, `k_async_db`, and the per-worker pg_async pool (identical pg_async footprint to this branch's head). The original commit hashes aren't in main because of the squash, but every line is. Nothing left to merge.

PR #888, opened 2026-06-17
vanilla-epoll: cross-request pipelining for async-db (stacked on #884)

enghitalo:
Wires the framework onto vanilla's **cross-request pipelining** (vanilla PR enghitalo/vanilla#45) — the fix for the per-worker pool starvation behind #884's async-db/fortunes regression.

**Stacked on #884** (`perf/vanilla-async-db-flatstate`, the async park/resume baseline). The net-new here is the pipelining wiring:
- `park()` picks the **least-loaded** connection (`acquire_pipelined`) and sheds only when every conn is at the `max_inflight`=8 cap — so a connection now carries up to N in-flight queries (per-worker DB concurrency `conns × N`), instead of the old one-in-flight-per-conn that capped it at 2 conns/worker and shed the overflow as empty 200s.
- `async_submit`'s shed-on-full bool is checked; `on_db_ready` drops the exclusive release. The reactor's per-fd watch queue fans each reply to its request in submission order.
- async-db / fortunes / crud-list all flow through `park()`, so all gain the concurrency.

## CI note
The Dockerfile is **temporarily pinned to vanilla `feat/pg-async-pipelining`** (vanilla#45) so this builds against the pipelining lib and can benchmark before #45 merges — revert to `refs/heads/main` once it lands.

## Validated
Driver against real PG18 (3-deep FIFO order + mid-pipeline error isolation), reactor unit-tested (promotion/FIFO/tombstone) + the async_pipe e2e, framework compiles against the pipelining lib. The async-db benchmark here is the end-to-end gate — it drives the concurrent parks → the reactor's multi-client drain under load (correct, non-empty results = the multi-drain works; throughput ↑ at lower CPU = the lever lands).


enghitalo:
Added one commit: **`perf(vanilla-epoll): arm pooled PG conns with watch_persistent`**.

This is a robustness fix for the pooled-connection path this PR introduces. Previously, a client disconnecting **mid-query** (while parked alone on a pooled PG connection) caused the runtime to **close that pooled connection** — forcing a reconnect + a fresh SCRAM/PBKDF2 auth handshake on the next borrow, a needless storm under disconnect-heavy load. The query handler now parks via `ac.watch_persistent`, so on a client disconnect the runtime drains the orphaned reply in order and keeps the connection **open for reuse** instead of closing it.

Depends on `watch_persistent` in the vanilla library (enghitalo/vanilla#47), which is now **merged to vanilla `main`** — so this builds against the `main` the framework Dockerfile clones. Verified locally: framework compiles against merged vanilla `main`; the library side is unit-tested + gated against PG18.

Ready for a re-`/benchmark` (async-db / api-* profiles) whenever convenient.

MDA2AV (approved):

PR #890, opened 2026-06-19
[C#] GenHTTP: Test effect of backlog

Kaliumhexacyanoferrat:
Test PR to compare the effect of different backlog settings. Currently defaults to 1024.

Do not merge.


Kaliumhexacyanoferrat:
So basically no effect at all.

PR #894, opened 2026-06-19
perf(vanilla-epoll): zero-alloc crud-write parser + upload streaming fix + V bump to cef604a26 (consolidates #935)

enghitalo:
## vanilla-epoll: zero-alloc crud-write parser **+** upload streaming fix **+** V toolchain bump (consolidates #935)

Three changes, one PR (per review — #935 closed into this):

1. **`parse_crud_body_fast`** — a zero-alloc JSON body parser for the crud write path (`start_crud_create` / `start_crud_update`), with a `json.decode` fallback. Borrows the field bytes straight out of the request buffer (no per-request decode allocation under `-gc none`).
2. **vanilla library pin `1822cbb` → `10488d2`** — picks up enghitalo/vanilla#61: the epoll large-body drain is fixed to *drain-then-respond*. Previously the drain was dead code (an inverted `flush_batch` answered `/upload` from the head and closed the connection without ingesting the body); it now streams the body off the socket and keeps the connection alive. `/upload` correctly ingests 500K/2M/10M/20M (validate green). See MDA2AV/HttpArena#939 for the harness gap that let the no-ingest behaviour score #1. Verified leak-free under `-gc none` (RSS flat under sustained uploads).
3. **V toolchain bump `c0624b274` → `cef604a26`** — required: the stale `c0624b274` pin inherited from `main` **crashes the server mid-validation** (it passed all baseline tests then died before `/pipeline`). On `cef604a26` the full validation runs cleanly locally (baseline + 4 TCP-fragmentation cases, pipelined, json, json-comp, upload 2M, static + 404, async-db, fortunes, and the crud write path: POST 201, PUT 200, X-Cache MISS→HIT, invalidate-on-PUT). The crash was a `c0624b274`-era codegen bug, already fixed upstream.

### Known CI caveat (pre-existing, not from this PR)
The `validate (vanilla-epoll)` job is intermittently flaky on the CI runner: `async-db`/`fortunes` (both on the `pg_async` park/resume path) occasionally hang (a lost-wakeup race that manifests on few-core runners). The same flake hit **#935** (main + V bump, no code from this PR) on `async-db`. It is **not reproducible locally** even pinned to 2 cores under heavy concurrent load (52K async-db RPS, 22K fortunes RPS, zero hangs/bad bodies). This needs a dedicated `pg_async` reactor investigation in the vanilla library; it is orthogonal to this PR's changes.


enghitalo:
@MDA2AV Can you please, merge this one?

MDA2AV (approved):

PR #902, opened 2026-06-20
vanilla-io_uring: modernize to V master (c0624b274) — Go-style db.pg + pinned vanilla lib

enghitalo:
Modernizes `vanilla-io_uring` to V master `c0624b274`, the last of the V-frameworks series ([fasthttp #898](https://github.com/MDA2AV/HttpArena/pull/898), [veb #900](https://github.com/MDA2AV/HttpArena/pull/900), [pico #901](https://github.com/MDA2AV/HttpArena/pull/901)).

- **`db.pg` → Go-style pooled API**, at every site (crud list/get/create/update, fortunes, async-db): `ConnectionPool` + manual `acquire()`/`release()` → `pg.connect(cfg, PoolConfig{max_open_conns: N})` + `db.exec_param_many()` directly (transparent pooled acquire/release; thread-safe). The async-db endpoint's old per-conn lazily-**prepared** statement is replaced with `exec_param_many` — prepared statements are session-scoped and the new pool hands out a transient conn per call, so a clean prepared-statement path needs `db.conn()` pinning (left as a follow-up; the re-parse cost is minor and this engine was untuned).
- **Default GC (`v -prod`), not `-gc none` and not `-prealloc`** — this runs on vanilla's own manual buffer management, not the stdlib veb/fasthttp per-request arena.
- **Dockerfile**: pin V at master `c0624b274` (the codegen regression vlang/v#27468 that forced the 0.5.1 pin is fixed in #27470); **and pin the vanilla library to a commit** instead of the `ADD <github-api-ref>` cache-bust that broke the build on a transient 504 (MDA2AV/HttpArena#895).
- README + meta.json updated.

Compile-verified with `v -prod .` (liburing-dev present). Test subscriptions unchanged. Runtime (io_uring + live Postgres) not exercised here.


enghitalo:
**Converting to draft / holding this PR.**

The modernization (V c0624b2 + warm `db.pg` pool) is a clear win on the profiles the io_uring engine is actually good at:

- **crud +138%** (41K → 98K RPS — the `max_idle_conns = max_open_conns` warm-pool fix)
- **baseline +25%**, **json-comp +10%**, **limited-conn +11%**, pipelined +2-3%

But the DB profiles are down vs the prior baseline and **can't be fixed at the app level**:

- **async-db −50%, fortunes −89%.** vanilla-io_uring uses the **blocking `db.pg`** driver on the io_uring loop. The epoll variant reaches async-db ~289K / fortunes ~167K (50-500× faster) *only* because it uses the non-blocking `pg_async` bridge, which needs the per-worker async runtime + per-worker state. The io_uring server exposes a single shared handler closure with **no per-worker hook**, so it can neither pin a connection per worker nor run async PG. The Go-style shared pool's per-call acquire/release is also ~2× the old `ConnectionPool`.
- **static-4096 / api** at high connection counts are a pre-existing io_uring-backend characteristic (static-4096 was already ~15K on 0.5.1, vs ~237K at 1024 conns), not introduced here.

**Held pending vanilla-lib work**: per-worker state on the io_uring backend + async PG (pg_async) there, so the DB profiles aren't capped by a blocking driver. Until then this PR isn't competitive on DB and shouldn't merge.

MDA2AV (approved):

PR #918, opened 2026-06-23
fasthttp: pin V to latest + cut per-request allocation (default GC)

### LINKED ISSUE #27468: Array single-element push (`<<`) 4-7x slower on master vs 0.5.1 (alloc, push_many, indexed-write unaffected)
## Describe the bug

Single-element array push (`arr << x`) is **4–7x slower on current `master`** than on the `0.5.1` release. Allocation, bulk `push_many`, and pre-sized indexed writes (`arr[i] = x`) are **all unaffected** (master is even marginally faster on those), so the regression is isolated to the per-element push path.

This also appears to be the **root cause of the `x.json2` encode regression** reported in #27467 — `x.json2` builds its output buffer via single-element pushes.

## Reproduction

Self-contained, no deps:

```v
import time

@[direct_array_access]
fn main() {
	n := 3_000_000
	mut sink := 0

	mut sw := time.new_stopwatch()
	for _ in 0 .. n {
		a := []int{len: 256, cap: 256}
		sink += a.len
	}
	alloc_ms := sw.elapsed().milliseconds()

	mut src := []u8{len: 256}
	sw = time.new_stopwatch()
	for _ in 0 .. n {
		mut out := []u8{cap: 256}
		unsafe { out.push_many(src.data, src.len) }
		sink += out.len
	}
	pushmany_ms := sw.elapsed().milliseconds()

	sw = time.new_stopwatch()
	for _ in 0 .. n {
		mut out := []u8{cap: 256}
		for i in 0 .. 256 { out << u8(i) }
		sink += out.len
	}
	push_ms := sw.elapsed().milliseconds()

	sw = time.new_stopwatch()
	for _ in 0 .. n {
		mut out := []u8{len: 256, cap: 256}
		for i in 0 .. 256 { out[i] = u8(i) }
		sink += int(out[255])
	}
	idx_ms := sw.elapsed().milliseconds()

	println('alloc=${alloc_ms}ms  push_many=${pushmany_ms}ms  single<<=${push_ms}ms  indexed=${idx_ms}ms')
}
```

Build with `v -prod` and run on each version.

## Expected vs actual

`v -prod`, gcc 14.2.0 (`-O3`), Linux x86_64:

| step (3M iters × 256) | V 0.5.1 (`0c3183c`) | V master (`065a450`) | ratio |
|---|---|---|---|
| allocation `[]int{len:256}` | 528 ms | 426 ms | 0.8× (fine) |
| `push_many` (bulk) | 141 ms | 114 ms | 0.8× (fine) |
| **single `<<`** | **421 ms** | **3161 ms** | **7.5× slower** |
| pre-sized `out[i] = x` | 122 ms | 106 ms | 0.9× (fine) |

A separate `int`-element variant shows ~4× and a `u8` variant ~5×, so the factor depends on element/workload, but single-element `<<` is consistently several times slower while every other array operation is unaffected.

## Likely cause

`master` has `builtin: guard array push len panics (#27335)` and related changes to the push path; the per-push guard added since 0.5.1 looks like the prime suspect. A `git bisect` between `0.5.1` and `065a450` on the repro above should localize it quickly.

## Environment

- **V master:** `065a450` (`V 0.5.1 7bf483e...065a450`)
- **V release for comparison:** `0.5.1` (`0c3183c`, release `v_linux.zip`)
- OS: Linux 6.8.0-41-generic, x86_64; cc: gcc 14.2.0 (`-prod` → `-O3`)

## Impact

`arr << x` in hot loops is extremely common (string/byte buffer building, serializers). This single regression cascades broadly — e.g. it is almost certainly the underlying cause of the `x.json2` ~7× encode regression in #27467. Workaround for hot paths: pre-size the buffer and use indexed writes (`out[i] = x`), which is unaffected and ~10× faster than `<<` on master.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

enghitalo:
## Summary
Pin the V toolchain used by frameworks/fasthttp from c0624b274 to the latest upstream V master commit 6c4d26f1f for reproducible and up-to-date benchmark builds.

## Why this can improve benchmark results
This update pulls in recent upstream V changes that are directly relevant to runtime behavior under load:

- Merged PR: https://github.com/vlang/v/pull/27544
  - cheaper slice marking and scalable GC can reduce GC and memory-management overhead in allocation-heavy HTTP workloads.
- Merged PR: https://github.com/vlang/v/pull/27470
  - Fixes slower single-element array push paths; this closed https://github.com/vlang/v/issues/27468.
  - This can help request/response code paths that append to small buffers repeatedly.
- Merged PR: https://github.com/vlang/v/pull/27265
  - db.pg moved to a thread-safe pool-backed API.
  - For benchmark scenarios using PostgreSQL-backed routes, this should reduce contention and improve concurrency behavior.

## Closed issue references
- Closed issue: https://github.com/vlang/v/issues/27468 (fixed by PR #27470)
- Closed issue: https://github.com/vlang/v/issues/27456 (addressed by https://github.com/vlang/v/pull/27490; primarily correctness/synchronization stability under concurrency)

## Scope
- Updated only: frameworks/fasthttp/Dockerfile

---

## Update: also reduce per-request allocation (default GC, no `-gc none`)

Following the GC study (`V_GC_TLA_STUDY.md`), fasthttp's `json` path is
GC-collection-bound, so fewer per-request allocations = less Boehm stop-the-world =
more throughput, with bounded memory. Three **output-identical** changes:

- `content_owned: true` on every response → the backend *takes* the response builder's
  buffer instead of cloning it (`take_or_clone_content`), removing one alloc per response.
- zero-copy routing: method/target/route are `tos()` views into the request buffer
  instead of `bytestr()` / `all_before()` copies.
- `parse_u_at` for the `/json/{count}` segment instead of `route[6..].i64()` (no substring).

Functionally validated end-to-end against Postgres (pipeline / baseline / json / async-db
/ upload). No `-gc none`.


enghitalo:
### Update: bumped pinned V to `cd213a17c` (picks up the #27555 GC fix)

The earlier pin `6c4d26f1f` sat inside the `c0624b2..6c4d26f` range that [vlang/v#27555](https://github.com/vlang/v/issues/27555) identified as a default-Boehm-GC throughput regression on multi-threaded, allocation-light HTTP workloads (json −36.7%, static −34–52%, RSS +24–38%).

That issue is now fixed by [vlang/v#27560](https://github.com/vlang/v/pull/27560) ("cgen: lower Boehm GC free-space divisor to 1 for the opt modes"), which is master commit `cd213a17c`. This branch is now pinned there so the rebuild measures the **recovered** default-GC throughput rather than the regressed range.

Re-run `/benchmark -f fasthttp` to capture the post-fix numbers.

MDA2AV (approved):

PR #922, opened 2026-06-23
vanilla-io_uring: pin V to latest + zero-alloc crud responses (default GC)

### LINKED ISSUE #27468: Array single-element push (`<<`) 4-7x slower on master vs 0.5.1 (alloc, push_many, indexed-write unaffected)

enghitalo:
## Summary
Pin the V toolchain used by frameworks/vanilla-io_uring from c0624b274 to the latest upstream V master commit 6c4d26f1f.

## Why this can improve benchmark results
This V bump brings in upstream merges that should help allocation-heavy and high-concurrency benchmark paths:

- Merged PR: https://github.com/vlang/v/pull/27544
  - cheaper slice marking and scalable GC reduce memory-management overhead and can improve tail latency under load.
- Merged PR: https://github.com/vlang/v/pull/27470
  - fixes a single-element array push regression and closes https://github.com/vlang/v/issues/27468.
  - Relevant for frequent small append operations in hot request paths.
- Merged PR: https://github.com/vlang/v/pull/27265
  - db.pg is now thread-safe and pool-backed, improving concurrency for PostgreSQL-backed handlers.

## Closed issue references
- Closed issue: https://github.com/vlang/v/issues/27468 (fixed by PR #27470)
- Closed issue: https://github.com/vlang/v/issues/27456 (addressed by https://github.com/vlang/v/pull/27490)

## Scope
- Updated only: frameworks/vanilla-io_uring/Dockerfile

---

## Update: zero-alloc crud response assembly (default GC, no `-gc none`)

The crud cache-HIT path runs at ~195K RPS, but `ok_xcache()` built a throwaway
`strings.Builder` per hit and the caller then copied it into `out` — pure GC pressure.
The crud handlers (get/list/create/update) now write the response **straight into the
connection buffer `out`** (via the existing `ws`/`wi` + a new `write_ok_xcache`/`wb`),
so the cache-HIT path allocates nothing and the others drop the response-head builder
and the body→out copy. **Byte-identical output**, validated against Postgres (X-Cache
MISS→HIT, 404, 201, invalidate-on-PUT, async-db, fortunes). No `-gc none`.

Note: this reduces GC pressure and memory; the `async-db`/`fortunes` throughput ceiling
is still architectural (blocking `db.pg` on the io_uring loop) and is tracked separately
— that part is unaffected by this change.


enghitalo:
### Update: bumped pinned V to `cd213a17c` (picks up the #27555 GC fix)

The earlier pin `6c4d26f1f` sat inside the `c0624b2..6c4d26f` range that [vlang/v#27555](https://github.com/vlang/v/issues/27555) identified as a default-Boehm-GC throughput regression — including the reproducible io_uring fortunes collapse (−77%).

That issue is now fixed by [vlang/v#27560](https://github.com/vlang/v/pull/27560) ("cgen: lower Boehm GC free-space divisor to 1 for the opt modes"), master commit `cd213a17c`. This branch is now pinned there so the rebuild measures the **recovered** default-GC throughput.

Re-run `/benchmark -f vanilla-io_uring` to capture the post-fix numbers.

MDA2AV (approved):

PR #930, opened 2026-06-26
Update validations, static tests for caching, refactor scripts

### LINKED ISSUE #927: Comparing apples to oranges
The default filter for https://www.http-arena.com shows Flagship and Emerging frameworks.

The newly added mq-bridge framework is marked as Emerging.
mq-bridge generates and caches response for the first request for the static calls.
Subsequent request return these cached responses.

Comparing frameworks that cache responses to frameworks that generate responses for each request doesn't seem fair.
A casual visitor won't know the difference between Flagship and Emerging and draw incorrect conclusions.

MDA2AV:


MDA2AV:
## Frameworks failing the new `static freshness (no pre-caching)` check

From a full local validation run on this branch across the static-supporting, non-engine frameworks. The new check mutates a static asset on disk mid-run and requires the response to reflect the change within 30s — i.e. **permanent in-memory preloading/caching of static files is no longer allowed** (standard or tuned).

**19 frameworks fail the check** (33 pass it, so it discriminates correctly rather than over-firing).

### Fail **only** this check — pure pass→fail flips introduced by this PR (13)
- `aspnet-minimal`
- `aspnet-minimal-aot`
- `aspnet-minimal-ioxide`
- `fletch`
- `frankenphp-trueasync`
- `genhttp-kestrel`
- `humming-bird`
- `rage`
- `rails`
- `servicestack`
- `swoole`
- `trillium-tuned`
- `userver`

### Fail this check **plus** an unrelated check (6)
- `genhttp` — also HTTP fragmentation (split-request returns empty body)
- `mq-bridge` — also `async-db` (returns `count=0`)
- `mq-bridge-py` — also `async-db` (returns `count=0`)
- `pyronova` — also `async-db`/`crud` (all empty)
- `quarkus-jvm` — also `async-db` (returns `count=0`)
- `ultimate-express` — also `json-comp` (no `Content-Encoding`)

Typical failure line:

```
FAIL [static freshness (no pre-caching)]: static response never reflected the on-disk
change within 30s — permanent pre-loading/caching of static files is not allowed
(standard or tuned).
```

PR #75, opened 2026-03-18
Add Rocket: type-safe Rust web framework (~26k ⭐)

BennyFranciscus:
## Rocket — type-safe Rust web framework

[Rocket](https://github.com/SergioBenitez/Rocket) is one of the most popular Rust web frameworks (~26k stars). Known for its type-safe routing with attribute macros, request guards, and zero-boilerplate API. Built on Tokio with async/await.

### Why this is interesting

HttpArena already has **actix** (actor-based), **axum** (extractor-based), **hyper** (raw), and **salvo** (feature-rich) — Rocket completes the Rust framework lineup. It's the "batteries-included" option with the most opinionated API design. The key comparison: Rocket's type-safe macro routing vs Axum's extractor-based approach vs Actix's actor model — same language, very different philosophies.

### Implementation

- Rocket 0.5 on Tokio, Rustls for TLS/H2
- Dual server: HTTP on 8080, HTTPS on 8443
- Custom `ServerResponse` responder with pre-set `Server: rocket` header
- Pre-computed gzip cache for compression endpoint
- Round-robin SQLite connection pool with mmap
- All standard endpoints: /pipeline, /baseline11, /baseline2, /json, /compression, /db, /upload, /static/{filename}
- `-O3`, thin LTO, single codegen unit

### Tests enabled

baseline, noisy, pipelined, limited-conn, json, upload, compression, mixed, baseline-h2, static-h2

cc @SergioBenitez — thought it'd be cool to see how Rocket stacks up against actix, axum, hyper, and salvo in HttpArena!

PR #951, opened 2026-06-29
perf(vanilla-io_uring): serve /static via static_assets (precompression + borrowed send)

enghitalo:
## What changed since the original

The original PR did a per-entry `core.queue_buf` borrowed send for the (identity-only) preloaded asset. This **supersedes** it: `/static` now goes through the lib's audited **`static_assets`** module, mounted at `/static/` (enghitalo/vanilla#80 `url_prefix`), which both **negotiates the precompressed `.br`/`.gz` sibling** AND emits small assets via **`core.queue_buf` borrowed send** — the #75 mechanism, folded into the module (enghitalo/vanilla#81). One audited static path, shared with vanilla-epoll (#954).

## Why

The `/static` handler preloaded **identity only** (skipping `.br`/`.gz`) and ignored `Accept-Encoding`, so it shipped the uncompressed body while the profile sends `Accept-Encoding: br;q=1, gzip;q=0.8`. Two wins now instead of one:

1. **precompression** — ships the ~4× smaller `.br` body (the dominant fix; vanilla-io_uring currently collapses on `static`: ~3.5K rps / 213 MB at static-6800);
2. **borrowed direct send** — the preloaded, immutable bytes are sent by the worker directly (`core.queue_buf`), never copied through the per-connection write buffer (the original PR's win, kept).

Drops the hand-rolled `StaticFile` / `static_response` / `content_type`.

## Verification

Built clean (`v -prod`). The io_uring + `static_assets` + `queue_buf` path verified **DB-free** on the io_uring backend (the entry itself needs Postgres at boot, so this used a minimal harness on the same lib): `/static/app.js` + `Accept-Encoding: br` → `Content-Encoding: br`, **47,275 B** (vs 204,800 identity); `components.css.br` delivered **byte-exact** (33,268 B). For the epoll sibling the same module gives **+83% rps / −52% bandwidth** in a 2-core A/B (#954) — io_uring should see a larger jump since it starts from the broken-static baseline.

## Pin

Bumps the vanilla pin to **main + #80 (url_prefix) + #81 (queue_buf emit)**. Depends on those two (both open against enghitalo/vanilla; #80 already merged). The Dockerfile pins the #81 branch commit so CI builds today; it lands on vanilla `main` once #81 merges.

### DIFF
--- a/frameworks/vanilla-io_uring/Dockerfile
+++ b/frameworks/vanilla-io_uring/Dockerfile
@@ -24,8 +24,8 @@ RUN git clone https://github.com/vlang/v /opt/v && \
 # main moved, but that hit api.github.com every build and a transient 504/rate-limit
 # failed the whole build (MDA2AV/HttpArena#895). Bump this commit to pick up lib fixes.
 #
-# vanilla bumped to main + the static_assets static path (enghitalo/vanilla#80
-# url_prefix mount, #81 core.queue_buf borrowed send for small assets): this entry
+# vanilla main @b189036 (incl. enghitalo/vanilla#80 url_prefix mount + #81
+# core.queue_buf borrowed send for small assets): this entry
 # now serves /static/* through the audited static_assets module — precompressed
 # .br/.gz negotiation per Accept-Encoding + ETag/Vary, with small assets emitted via
 # core.queue_buf borrowed send (the #75 mechanism, now folded into the module) so the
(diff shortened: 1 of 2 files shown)


enghitalo:
Superseded by #957 — vanilla-io_uring static + json-tls consolidated into one PR to avoid the entry/Dockerfile conflict.

PR #952, opened 2026-06-29
feat(vanilla-epoll): add json-tls profile (HTTP/1.1 over TLS 1.3)

enghitalo:
## What

Wires the **json-tls** profile for `vanilla-epoll`: the existing `/json/{count}?m={m}` endpoint served over **HTTP/1.1 + TLS 1.3** on `:8081`. This was the only gap keeping vanilla off the json-tls board — the JSON path already existed and is correct; this PR adds the TLS transport and declares the test.

The JSON serializer is reused **verbatim**, so json-tls responses are byte-for-byte identical to the plaintext `/json` (verified).

## How

**`main.v`**
- Extract a transport-agnostic `write_json_into(ro, mut out, count, m)`; the plaintext `WorkerCtx` method now delegates (no behavior change). Content-Length is precomputed from the same values the body emits → the framed length can never desync (no response-splitting surface).
- Add a **stateless** TLS `request_handler` that serves **only `/json`** (404 elsewhere), keeping the TLS port's surface minimal. It captures the read-only `ro`; no `make_state` (sidesteps the TLS worker's stateful path).
- Start a second `http_server` on `TLS_PORT` (env, default `8081`) with `tls_config` from `/certs/server.{crt,key}` (`TLS_CERT`/`TLS_KEY` overridable). Fails loud if the cert is present but the key is missing; self-signs only when no cert is mounted.

**`Dockerfile`**
- Build **Mbed TLS 4.1.0** from the pinned release tarball; compile with `-d vanilla_tls`. Runtime carries the shared libs (incl. `libtfpsacrypto`, the 4.x TF-PSA-Crypto split) + `ldconfig`. `EXPOSE 8081`.
- **Thread-safety (load-bearing):** enable `MBEDTLS_THREADING_C` + `MBEDTLS_THREADING_PTHREAD`. vanilla runs N TLS worker threads, all driving TLS 1.3 handshakes through Mbed TLS's **global PSA key store**; without the threading mutex the concurrent handshakes race on key slots → a **heap-use-after-free** under load (ASan: `psa_wipe_key_slot` frees a slot another thread is mid-`memcpy` in `psa_hmac_setup`; crashes ~c1024 with `double free or corruption`). With it enabled, the load that 100%-crashed now survives clean.

**`meta.json`**: add `"json-tls"` to `tests`.

## Validation

Built the image and ran it with the repo cert bind-mounted at `/certs`:

- ✅ All 3 `validate.sh` json-tls assertions: protocol negotiation (HTTP/1.1, ALPN `http/1.1` over TLS 1.3), response schema + `total == price*quantity*m` on `7:2 / 23:11 / 50:1`, `Content-Type: application/json`.
- ✅ Byte-for-byte parity: TLS `/json` output identical to plaintext `/json`.
- ✅ Keep-alive (curl `num_connects=1`, connection reused) — the profile's defining property.
- ✅ No heap corruption under concurrent TLS load after the threading fix (in-container c1024 ~137k rps, c1536 ok; clean log).

> Note: local numbers are from a contended dev box (loopback) and are **not** leaderboard-comparable — they're stability/correctness signals. A leaderboard run belongs on the dedicated bench host.


enghitalo:
## Update: json-tls now runs over kTLS (kernel record offload)

Pushed a follow-up that bumps the vanilla lib pin `15bd57e → 7ca36f6` ([`perf/ktls-record-offload`](https://github.com/enghitalo/vanilla/tree/perf/ktls-record-offload)). After the Mbed TLS handshake the connection hands record crypto to the **kernel** (`setsockopt TCP_ULP "tls"` + `TLS_TX`/`TLS_RX`, AES-128-GCM), so the epoll TLS worker's steady-state read/write become plain `recv()`/`send()`.

**Why:** the prior userspace run landed at **512k rps** (~76% TLS tax). Root cause: Mbed TLS 4's PSA crypto uses a process-global key store, and `MBEDTLS_THREADING_C` (required for vanilla's N TLS worker threads) wraps it in a mutex that serializes **every record** across workers. kTLS removes all per-record userspace crypto — and that mutex — from the hot path; the handshake still uses Mbed TLS (amortized by keep-alive). The suite is pinned to `TLS_AES_128_GCM_SHA256` and TLS 1.3 tickets disabled so the kernel record sequence starts at 0.

**Clean A/B:** the 5 changed TLS files are byte-identical at `15bd57e` and lib HEAD, so this pin is exactly `15bd57e` + kTLS.

**Local validation (this dev box has no `tls` kernel module + no sudo, so kTLS can't engage here — only the bench host can):**
- HKDF-Expand-Label key derivation **cross-checked byte-identical against `openssl kdf HKDF`**; the `tls12_crypto_info_aes_gcm_128` layout / IV-salt split / direction (server→TX, client→RX) / `rec_seq=0` verified against `linux/tls.h` + RFC 8446 §7.
- Integration + **clean fallback**: handshake → try kTLS → (`TCP_ULP` ENOENT here) → userspace Mbed TLS → correct responses (3 validate pairs pass, ALPN `http/1.1`, cipher pinned to AES-128-GCM, c512 stable). So json-tls **never breaks** on a host without the module.

A one-time stderr line logs `[ktls] engaged: kernel TLS TX+RX` (or `[ktls] fallback: ...`) per process — on the bench host it should read **engaged**. The benchmark there is what validates the perf (target: top the json-tls board, currently zix @ 1.90M).

enghitalo:
## kTLS result: json-tls **512k → 1,502,783 rps** (2.93×)

The kTLS pin engaged on the bench host — json-tls jumped from the userspace **512,243** to **1,502,783 rps**, cutting the TLS tax from ~76% to ~30% (plain json is 2.15M) at slightly *lower* CPU (6163% vs 6473%). That lands vanilla-epoll in the json-tls top tier (just under swerver's 1.52M), up from mid-pack.

Also note `async-db +8.4%` / `crud +5.3%` this run — the `crud -14.6%` in the userspace run was just variance (the TLS listener is idle during non-TLS profiles).

This PR is now a solid, mergeable result: correct json-tls (validate passes), fast via kTLS where the `tls` module is present, and a **clean userspace fallback** where it isn't (so it never breaks). 

Still ~21% behind the io_uring leaders (zix 1.90M / sark 1.83M / ioxide 1.79M). The remaining gap is structural, not crypto: epoll's single-acceptor + per-edge syscalls vs io_uring multishot/batched, plus the 4096-handshake startup storm still going through Mbed TLS's global PSA mutex. Closing it (SO_REUSEPORT per-worker accept to parallelize the handshake storm + accept) is a separate follow-up.

enghitalo:
@MDA2AV could you please merge this one?

enghitalo:
Superseded by #956 — vanilla-epoll json-tls consolidated with the static_assets change into one PR (the two conflicted on main.v + Dockerfile).

PR #954, opened 2026-06-29
perf(vanilla-epoll): serve /static via static_assets (precompressed negotiation)

enghitalo:
## Problem

`vanilla-epoll` loses the `static` profile by a wide margin — and the bandwidth column shows exactly why:

| static-4096 | rps | bandwidth |
|---|---:|---:|
| zix | 2.04M | 31 GB |
| **vanilla-epoll** | **1.15M** | **68 GB** |

~2× the bandwidth at ~half the rps. The `/static` handler preloaded **only the identity files** (it skipped the `.br`/`.gz` siblings) and never read `Accept-Encoding`, so it always shipped the **uncompressed** body. The profile sends `Accept-Encoding: br;q=1, gzip;q=0.8`, so the leaders serve the ~4× smaller `.br` sibling and vanilla was sending the full file.

It was also a **second, hand-rolled static path** — the lib already ships an audited `static_assets` module that does precompressed negotiation + ETag/conditional + `Vary` + `sendfile(2)` correctly, but the entry didn't use it.

## Change

Serve `/static/*` through `static_assets`, mounted at `/static/` via the new `url_prefix` (depends on **enghitalo/vanilla#80**). One audited implementation; drops the duplicate `StaticFile` / `static_header` / `content_type` / `C.open`. Emits via the same `core.queue_file` sendfile handoff the worker already drains.

## Evidence

Local A/B, server pinned to **2 cores**, `wrk` + this repo's `static-rotate.lua` (`Accept-Encoding: br;q=1`), both CPU-saturated:

| `/static` handler | req/s | Transfer/s |
|---|---:|---:|
| identity-only (current) | ~56.8K | 3.37 GB/s |
| **`static_assets` (this PR)** | **~104K** | 1.60 GB/s |

**+83% rps, −52% bandwidth** at equal CPU — the precompressed body is the whole difference (mirrors the published zix-vs-vanilla static signature). curl-verified correctness: `/static/app.js` + `Accept-Encoding: br` → `Content-Encoding: br`, 47,275 B (vs 204,800 identity); `.woff2` (no sibling) → identity; `..` traversal → 404; `gzip`-only → `.gz`.

## Pin

Bumps the vanilla pin to **main + https://github.com/enghitalo/vanilla/pull/80** (`url_prefix`). https://github.com/enghitalo/vanilla/pull/75 (`core.queue_buf`) and https://github.com/enghitalo/vanilla/pull/79 (kTLS) ride along on main but are **inert** in this image: it builds without `-d vanilla_tls`, and `queue_buf` is unused by the epoll entry. Compiles clean (`v -prod -gc none`).

> Depends on enghitalo/vanilla#80 (the `url_prefix` field). The Dockerfile pins the branch commit so CI builds today; it lands on vanilla `main` once https://github.com/enghitalo/vanilla/pull/80 merges.

### DIFF
--- a/frameworks/vanilla-epoll/Dockerfile
+++ b/frameworks/vanilla-epoll/Dockerfile
@@ -22,14 +22,14 @@ RUN git clone https://github.com/vlang/v /opt/v && \
 # (or rate-limit) failed the whole build (MDA2AV/HttpArena#895). To pick up upstream
 # library fixes, bump this commit.
 #
-# vanilla bumped to main + enghitalo/vanilla#80 (static_assets `url_prefix`): this
-# entry now serves /static/* through the audited static_assets module (precompressed
-# .br/.gz negotiation per Accept-Encoding + ETag/Vary + sendfile) instead of a second,
-# hand-rolled identity-only map that ignored Accept-Encoding. The bump also catches up
-# to current vanilla main (#75 core.queue_buf, #79 kTLS record offload) — both inert
-# in this image (built without -d vanilla_tls; queue_buf is unused by the epoll entry).
+# vanilla main @b189036 (incl. enghitalo/vanilla#80 static_assets `url_prefix` and
+# #81 core.queue_buf borrowed send): this entry serves /static/* through the audited
+# static_assets module (precompressed .br/.gz negotiation per Accept-Encoding +
+# ETag/Vary + sendfile) instead of a second, hand-rolled identity-only map that
+# ignored Accept-Encoding. #81's borrowed send and #79 kTLS are inert here (built
+# without -d vanilla_tls; epoll serves small bodies via the write buffer).
 RUN git clone https://github.com/enghitalo/vanilla /root/.vmodules/vanilla && \
-    git -C /root/.vmodules/vanilla checkout f59a534045e5731a6c03f16ea23f19cbfe2e7334
+    git -C /root/.vmodules/vanilla checkout b189036212e4283ef2cffe42b318b556f8a3d1bc
 
 WORKDIR /app
 COPY . .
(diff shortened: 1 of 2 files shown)


enghitalo:
Dependency satisfied: enghitalo/vanilla#80 (`url_prefix`) is merged, so the pinned commit `f59a534` is now on vanilla `main`. (The borrowed-send emit for small assets — enghitalo/vanilla#81 — is **not** needed here: it's inert on epoll, which serves the body via `out << response` either way. It matters for the io_uring sibling, MDA2AV/HttpArena#951.)

enghitalo:
The benchmark bot caught a memory regression in the previous revision: switching `/static` to `static_assets` traded the old `sendfile` path for **preload + copy** on epoll — each connection copied its ~33–47 KiB `.br` response through the per-connection write buffer (which the pool retains), giving **+168% memory (~1 GiB at 6800 conns)** and a **−9.8% rps** regression at the top conn count, for only +7–24% at lower counts.

Pushed a fix: `sendfile_min_bytes: 16 KiB` so the large `.br` siblings (the balloon drivers) are served disk-backed via `sendfile(2)` (zero-copy, headers-only in the write buffer) while the tiny ones still preload (cheap memcpy). Local A/B (4 server cores, `wrk` static-rotate `br`, 512 conns):

| variant | RSS | rps |
|---|--:|--:|
| old (identity via sendfile) | 65 MiB | 120K |
| prev rev (preload+copy .br) | 70 MiB | 200K |
| **this fix (sendfile .br ≥16K)** | **13 MiB** | **207K** |
| sendfile-all | 6 MiB | 162K |

Flat memory **and** best throughput — strictly better than both preload-all and sendfile-all. io_uring (#951) is unaffected: it keeps preload + `core.queue_buf` borrowed send, which is already flat-memory (and was the big win there: static 3K→1.4M, mem −33%).

enghitalo:
Superseded by #956 — vanilla-epoll static + json-tls consolidated into one PR to avoid the entry/Dockerfile conflict between the static and json-tls changes.

PR #955, opened 2026-06-29
feat(vanilla-io_uring): add json-tls (epoll+kTLS TLS listener on :8081)

enghitalo:
## What

Adds the **json-tls** profile to `vanilla-io_uring`. The lib's io_uring backend has no TLS, so the json-tls listener runs on a **second server bound to :8081 using the lib's epoll backend** — the same TLS 1.3 + **kTLS** path that landed for `vanilla-epoll` (PR #952). The io_uring server keeps serving the non-TLS profiles on :8080. Same dual-server pattern.

After the Mbed TLS handshake the kernel does record AES-128-GCM via `TLS_TX`/`TLS_RX` where the `tls` module is present (clean userspace fallback otherwise), so the steady-state read/write are plain `recv`/`send` with no per-record userspace crypto.

## Why this shape

The io_uring backend would need a TLS handshake driven by io_uring completions — a sizable lib project. json-tls is a **separate port** (8081), so the pragmatic, low-risk move is to reuse the proven epoll+kTLS listener there and keep io_uring for the hot non-TLS paths. (Pure io_uring TLS is a possible future follow-up.)

## How

**`main.v`**
- A **stateless** `request_handler` on the :8081 epoll server serving **only** `/json` (404 elsewhere), reusing the existing allocation-free `write_json_response` (read-only: dataset + prefixes). It captures `sh` and never touches the DB/caches, so it runs safely alongside the io_uring workers. `TLS_PORT` (default 8081) for local runs; spawned before the blocking io_uring `server.run()`.
- `load_tls_config`: reads `/certs/server.{crt,key}` (`TLS_CERT`/`TLS_KEY` overridable); fails loud if the cert is present but the key isn't; self-signs when no cert is mounted.

**`Dockerfile`**
- Bump the vanilla lib pin to main `@b189036` (includes kTLS, enghitalo/vanilla#79).
- Build Mbed TLS 4.1.0 from source with `MBEDTLS_THREADING_C`/`PTHREAD` (the TLS workers share Mbed TLS's global PSA key store), compile with `-d vanilla_tls`, carry the shared libs at runtime + `ldconfig`, `EXPOSE 8081`.

**`meta.json`**: add `"json-tls"`.

## Validation (local; this box has no `tls` kernel module → userspace fallback)

- Both servers coexist: io_uring `:8080` + epoll-TLS `:8081`. Plaintext `/json` on 8080 → 200.
- The 3 json-tls validate pairs (7:2 / 23:11 / 50:1) pass — item schema + `total == price*quantity*m`.
- ALPN `http/1.1`, cipher pinned to `TLS_AES_128_GCM_SHA256`, one-time `[ktls] fallback` log fires (no module here).

On the bench host kTLS engages — the same epoll+kTLS code that took `vanilla-epoll` json-tls to **1.50M rps**. Expect a similar number here (the TLS listener is byte-identical; the difference vs vanilla-epoll is only the :8080 backend).

### DIFF
--- a/frameworks/vanilla-io_uring/Dockerfile
+++ b/frameworks/vanilla-io_uring/Dockerfile
@@ -2,7 +2,8 @@ FROM debian:stable-slim AS build
 
 RUN apt-get -qq update && \
     apt-get -qy install --no-install-recommends \
-        ca-certificates build-essential git libpq-dev liburing-dev && \
+        ca-certificates build-essential git libpq-dev liburing-dev \
+        cmake curl bzip2 python3 && \
     rm -rf /var/lib/apt/lists/*
 
 # Pinned, reproducible V at master commit 02015a7c (built from source). Master is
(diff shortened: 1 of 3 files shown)


enghitalo:
Superseded by #957 — vanilla-io_uring json-tls consolidated with the static_assets change into one PR.

PR #958, opened 2026-06-30
fix(vanilla-epoll/io_uring): lazy json-comp gz cache — drop ~135 MiB -gc none boot leak

### LINKED ISSUE #1: Metadata improvements to add more info

enghitalo:
## What

Restore the **lazy, shared, lock-guarded** json-comp gz cache on `vanilla-epoll` and `vanilla-io_uring`, reverting the lock-free boot precompute that shipped in #956/#957.

## Why

The precompute compressed the full `(count × m)` grid (~800 keys) at boot. On the **`-gc none`** `vanilla-epoll` build, V's `gzip.compress` leaks ~157 KiB of internal scratch per call (upstream bug, [vlang/v#27606](https://github.com/vlang/v/issues/27606)) — so the boot precompute leaked **~135 MiB of permanent RSS**. The last benchmark run showed it directly:

| vanilla-epoll mem | before (#956 combo) | after precompute |
|---|---|---|
| baseline 512 | 77 MiB | 213 MiB |
| json-comp 16384 | 221 MiB | 377 MiB |
| static 1024 | 100 MiB | 235 MiB |

(flat ~+135 MiB across *every* profile — the signature of a one-time boot allocation, reproduced exactly in isolation: 800 × `gzip.compress` under `-gc none` = ~135 MiB, reclaimed under default GC.)

The lazy cache compresses only the **8 `(count,m)` pairs the json-comp profile actually sends** (`/json/{1,5,10,15,25,40,50}` with `m∈1..8`), ≈ 1.3 MiB. This PR restores the **exact** json-comp code CI measured at 77 MiB in the #956/#957 *combo* commit, plus a guard comment so the precompute isn't reintroduced.

## The lock-free change had no upside

It was meant to fix the io_uring `json-comp@16384` collapse (a lock-contention hypothesis). The benchmark refuted it:

- **epoll** is healthy at 16384c **with** the shared lock (2.32M rps, CPU ~6100% — saturated).
- **io_uring** collapsed regardless, and *worse* with the precompute: 1.88M → **159K rps**, CPU 4312% → **438%**.

CPU *dropping* at max conns = rings starving, i.e. **worker/thread oversubscription** from the co-hosted json-tls listener (io_uring workers + the TLS thread pool exceed cores at 16384c), not lock contention. The real fix is a per-server worker cap ([enghitalo/vanilla#89](https://github.com/enghitalo/vanilla/issues/89) fix #1) — tracked as a follow-up, not in this PR.

`vanilla-io_uring` is default-GC so it didn't leak, but it's reverted too for parity and to restore its validated pre-cleanup state.

## Validation

- Diff is **only** the json-comp revert + an explanatory comment; both files `vfmt`-parse clean.
- The restored json-comp is byte-identical to the #956/#957 combo code that CI already measured at 77 MiB.
- Expected on a re-run: `vanilla-epoll` baseline mem back to ~77 MiB; json-comp behavior unchanged from the combo (healthy epoll@16384). No `/benchmark` triggered here — happy to run it on request.

### DIFF
--- a/frameworks/vanilla-epoll/main.v
+++ b/frameworks/vanilla-epoll/main.v
@@ -69,7 +69,8 @@ mut:
 	// (~25 MiB, lazily allocated).
 	crud    []CrudSlot
 	crud_mu &sync.RwMutex = unsafe { nil }
-	gz      map[u64][]u8 // json-comp: precomputed at boot, READ-ONLY during serving (lock-free reads)
+	gz      map[u64][]u8 // json-comp: (count<<32)|m -> gzipped response bytes
+	gz_mu   &sync.RwMutex = unsafe { nil }
 }
 
 // CrudSlot is one entry of the id-indexed crud cache slab. `buf` is the rendered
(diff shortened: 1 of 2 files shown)


enghitalo:
Superseded — split into one PR per engine as requested:
- **#959** — vanilla-epoll (the ~135 MiB `-gc none` memory fix)
- **#960** — vanilla-io_uring (parity; drops the pointless boot precompute)

Same changes, one file each. The benchmark results posted here already validated both engines (epoll 213→77 MiB, no regression; io_uring json-comp@16384 159K→402K). Closing in favor of #959 + #960.

PR #965, opened 2026-07-03
perf(vanilla-io_uring): converge onto the epoll twin — zero-alloc rendering, /pipeline skip-decode, crud slab (impl vanilla#84, #85)

enghitalo:
## What & why

`vanilla-io_uring` was an under-optimized copy of its `vanilla-epoll` twin: same handlers, but allocating throwaway strings per request, using `json.encode`/`json.decode` reflection on the DB paths, and parsing the full request even for the fixed `/pipeline` blit. This PR ports **every backend-agnostic optimization** from the epoll entry so the two share one audited set of response builders and diff cleanly — the stated goal of keeping the two entries easy to maintain in lock-step.

The io_uring backend of the vanilla library supports **only a stateless `request_handler`** (no `async_handler` / `make_state` / TLS — see enghitalo/vanilla#83, enghitalo/vanilla#93), so DB access necessarily stays on the blocking `db.pg` client. Everything that does **not** require the async runtime or per-worker state now matches epoll **byte-for-byte**.

Implements enghitalo/vanilla#84 (zero-alloc int parse/format) and enghitalo/vanilla#85 (crud: 1-query list, byte-rendered GET, fast crud-body parse).

## Changes (all entry-only, no lib change)

- **`wi` is now negative-aware** — fixes a latent wrong body for a negative `/baseline11` sum (`a=-10&b=3` returned garbage; now `-7`).
- **`emit` / `emit_int` (stack scratch) / `emit_xcache`** — zero-alloc response framing. `/baseline11` and `/upload` no longer allocate an `int -> string` per request (the highest-RPS non-DB profiles).
- **`/pipeline` skip-decode fast path** — blit the constant before any parsing; `decode_into` (no `!HttpRequest` Result boxing) on the main parse path.
- **`render_item_pg`** — byte-level JSON straight from `db.pg` text rows, removing the per-request `json.encode` reflection on `/async-db`, `/crud` list and `/crud` GET.
- **crud cache is now an id-indexed slab** (replaces `map[int]string`) with in-place buffer reuse and cache-aside invalidation, shared across ring workers under `RwMutex` — identical structure to the epoll twin.
- **`crud_list` uses a single windowed query** (`count(*) OVER()`) instead of a page `SELECT` + a separate `count(*)`.
- **`parse_crud_body_fast`** + borrowed JSON field parsers (with the `json.decode` fallback kept for escaped bodies).
- **`parse_i64_slice` / `dechunk_into` / `parse_hex_slice`** — allocation-free query/body parsing (`qint` no longer materializes a string per param).
- **static: `sendfile_min_bytes = 16 KiB`**, matching epoll (bounds per-connection RSS at high conn counts).

Net: the non-DB hot paths (pipeline / baseline / upload / json / json-comp / static) are now zero-alloc under the default GC, which should also pull io_uring's steady-state memory down toward epoll's (baseline was ~1.5 GiB vs epoll's ~78 MiB, dominated by the per-request `sum.str()` / `json.encode` churn this PR removes).

## What is intentionally NOT changed

DB profiles (`fortunes`, `async-db`, `api-*`, `crud`) stay capped by the **blocking `db.pg`** on the single ring worker — the io_uring backend has no async runtime to await DB readiness on the ring. Tracked in enghitalo/vanilla#83 (async runtime) and enghitalo/vanilla#93 (per-worker state). Those are the only remaining divergences from the epoll twin.

## Validation

- Both images build (`v -prod -d vanilla_tls`).
- Ran **both** containers against a **pristine seeded Postgres** (fresh DB per framework) and diffed every route: `pipeline`, `baseline11` (positive **and negative**), `upload`, `json`, `json-comp`, `async-db`, `fortunes`, `static` (br negotiation), `crud` list, `crud` GET (MISS then HIT), `crud` create, `crud` update, 404, and `json-tls` — **all 17 byte-for-byte identical to vanilla-epoll**.
- X-Cache sequence verified: `GET` MISS → HIT, then re-MISS after a `PUT` (slab invalidation). `POST` → 201. `json-comp` → `Content-Encoding: gzip`. `json-tls` → 200 over TLS 1.3.

### DIFF
--- a/frameworks/vanilla-io_uring/main.v
+++ b/frameworks/vanilla-io_uring/main.v
@@ -1134,16 +1134,21 @@ fn main() {
 
 	static_dir := os.getenv_opt('STATIC_DIR') or { '/data/static' }
 	// Canonical static server: loads every asset PLUS its .br/.gz siblings once, mounts
-	// them at /static/, and negotiates Accept-Encoding per request. sendfile_min_bytes
-	// matches the epoll twin: serve any representation >= 16 KiB from disk with sendfile
-	// instead of preloading + copying it through the per-connection write buffer (the
-	// large .br siblings otherwise balloon per-connection RSS at high conn counts). The
-	// tiny siblings stay preloaded (cheap memcpy).
+	// them at /static/, and negotiates Accept-Encoding per request, emitting via
+	// core.queue_buf (borrowed send of the preloaded bytes).
+	//
+	// Do NOT set sendfile_min_bytes here (unlike the epoll twin, which uses 16 KiB): the
+	// io_uring backend has NO sendfile path (no core.enable_sendfile / queue_file drain),
+	// so static_assets.respond_into would fall back to READING a "large" asset's body
+	// from disk on every request — a blocking read that stalls the ring. Keeping the
+	// default (256 KiB) preloads every arena .br/.gz sibling (all < 256 KiB) and serves
+	// them as a zero-copy borrowed send. (Measured: 16 KiB here collapsed static ~-86% to
+	// -99% — the large .br siblings hit the read-per-request fallback. See vanilla#93/#83:
+	// io_uring sendfile support is future work.)
 	asv := static_assets.new(static_assets.Config{
-		root:               static_dir
-		url_prefix:         '/static/'
-		spa_fallback:       ''
-		sendfile_min_bytes: 16 * 1024
+		root:         static_dir
+		url_prefix:   '/static/'
+		spa_fallback: ''
 	}) or { panic('vanilla-io_uring: static_assets init failed: ${err}') }
 
 	mut sh := Shared{
--- a/frameworks/vanilla-io_uring/main.v
+++ b/frameworks/vanilla-io_uring/main.v
@@ -28,42 +28,6 @@ struct DatasetItem {
 	rating   Rating
 }
 
-struct DbItem {
-	id       int
-	name     string
-	category string
-	price    int
-	quantity int
-	active   bool
-	tags     []string
-	rating   Rating
-}
-
-struct DbResp {
-	items []DbItem
-	count int
-}
-
-struct Fortune {
-	id      int
-	message string
-}
-
-struct Shared {
-mut:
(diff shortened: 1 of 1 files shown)


enghitalo:
Pushed a fix for a **static regression** the benchmark surfaced (static **-86% / -86% / -99%**, CPU collapsed): the converged entry had ported epoll's `sendfile_min_bytes = 16 KiB`, but the **io_uring backend has no sendfile path** (no `core.enable_sendfile` / `queue_file` drain). `static_assets` then served every representation ≥ 16 KiB (the large `.br`/`.gz` siblings) via a **per-request disk read** that stalls the single ring worker.

Fix (4bb52d0): drop `sendfile_min_bytes` on the io_uring entry → default 256 KiB → every arena sibling (< 256 KiB) is preloaded and served as a zero-copy `core.queue_buf` borrowed send. The epoll twin keeps `sendfile_min_bytes` because its backend *does* support sendfile.

Verified locally (wrk, 64 conns): `/static/vendor.js` (→ 67 KB `.br`) now serves at ~**101k req/s / 6.38 GB/s, zero socket errors** (preloaded, bandwidth-bound) instead of the disk-read collapse. Could you re-run the benchmark?

Re the other deltas in the run: the DB/api profiles (`async-db`, `api-16`, `crud`, `fortunes`) are all gated on the blocking-libpq single-ring-worker cap (enghitalo/vanilla#83), so at those small absolute rps they swing on DB/scheduling noise (e.g. `fortunes` +665%, `api-4` +9.8% but `api-16` -47% in the same run); `baseline`/`limited-conn` are engine-bound and io_uring-loopback-noisy (note `pipelined` went **+13.6%** from the /pipeline skip-decode, same code path as baseline). The one real item is the api memory growth (api-4 mem +125%) from the per-request DB-render scratch — tracked as a follow-up now that io_uring supports `make_state` (enghitalo/vanilla#93).

enghitalo:
Superseded by #967 — reopened clean with the static-config fix folded in (io_uring has no sendfile, so no `sendfile_min_bytes`) instead of a regression-then-fix history. Same convergence work.

PR #966, opened 2026-07-03
refactor(vanilla-epoll): extract shared emit_xcache helper for io_uring parity

enghitalo:
## What & why

Small maintainability refactor, paired with the `vanilla-io_uring` convergence PR. `vanilla-epoll` inlined the crud X-Cache response framing in two places (the `crud` GET cache **hit** and **miss** paths). This folds both into one `emit_xcache` helper.

The helper is **byte-identical** to the code it replaces, and **identical to the `emit_xcache` now used by the `vanilla-io_uring` twin** — so both entries share one audited X-Cache framing and diff cleanly. `vanilla-epoll` remains the performance reference; this only removes duplication.

## Change

- Add `emit_xcache(mut out, ctype, body, cache)` next to `emit` / `write_resp`.
- Use it in `start_crud_get` (HIT) and `render_crud_get` (MISS).

No behavior or performance change.

## Validation

Verified against a seeded Postgres that the `crud` GET **MISS** and **HIT** responses are byte-for-byte unchanged (and identical to the `vanilla-io_uring` twin, whose full-route byte-diff is covered in its PR).

### DIFF
--- a/frameworks/vanilla-epoll/main.v
+++ b/frameworks/vanilla-epoll/main.v
@@ -236,6 +236,21 @@ fn write_resp(mut out []u8, ctype string, body string) {
 	ws(mut out, body)
 }
 
+// emit_xcache writes a complete 200 JSON response carrying an X-Cache: HIT|MISS header
+// straight into `out` — the single framing shared by the crud GET hit/miss paths (and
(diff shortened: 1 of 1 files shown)


enghitalo:
Superseded by #968 — same `emit_xcache` refactor, reopened clean on current main.

Issue #94, opened 2026-03-24
Fix Json serializtion on cheating frameworks


MDA2AV:
14 frameworks are CHEATING — pre-serializing JSON at startup and returning
  cached bytes:

  1. blitz — dataset_json_resp pre-built at startup
  2. django — json_response_buf = json.dumps(...) at module load
  3. gleam-mist — ctx.json_cache pre-built via build_json_response()
  4. h2o — json_response pre-serialized with yajl at startup
  5. hono-bun — smallPayload pre-stringified at startup
  6. hummingbird — state.jsonCache from buildJsonCache()
  7. koa — jsonBuf = Buffer.from(JSON.stringify(...)) at startup
  8. ktor — AppData.jsonCache from buildJsonCache()
  9. phoenix — :persistent_term.put(:json_cache, ...) at startup
  10. rails — @@json_payload = JSON.generate(...) at class load
  11. sinatra — set :json_payload, JSON.generate(...) at configure
  12. starlette — json_buf = orjson.dumps(...) at module load
  13. vapor — state.jsonCache from buildJsonCache()
  14. vertx — jsonResponse = MAPPER.writeValueAsBytes(...) at startup

PR #967, opened 2026-07-04
perf(vanilla-io_uring): converge onto the epoll twin — zero-alloc rendering, /pipeline skip-decode, crud slab (impl vanilla#84, #85)

enghitalo:
## What & why

`vanilla-io_uring` was an under-optimized copy of its `vanilla-epoll` twin: same handlers, but allocating throwaway strings per request, using `json.encode`/`json.decode` reflection on the DB paths, and fully parsing the request even for the fixed `/pipeline` blit. This PR ports **every backend-agnostic optimization** from the epoll entry so the two share one audited set of response builders and diff cleanly.

The io_uring backend supports **only a stateless `request_handler`** (no `async_handler` / TLS — see enghitalo/vanilla#83), so DB access stays on the blocking `db.pg` client. Everything that does **not** require the async runtime now matches epoll **byte-for-byte**.

Implements enghitalo/vanilla#84 (zero-alloc int parse/format) and enghitalo/vanilla#85 (crud: 1-query list, byte-rendered GET, fast crud-body parse).

> Supersedes #965 (same work, rebased clean with the static fix folded in — no regression-then-fix history).

## Changes (all entry-only, no lib change)

- **`wi` is now negative-aware** — fixes a latent wrong body for a negative `/baseline11` sum (`a=-10&b=3` now returns `-7`).
- **`emit` / `emit_int` (stack scratch) / `emit_xcache`** — zero-alloc response framing. `/baseline11` and `/upload` no longer allocate an `int -> string` per request.
- **`/pipeline` skip-decode fast path** — blit the constant before any parsing; `decode_into` (no Result boxing) on the main parse path.
- **`render_item_pg`** — byte-level JSON straight from `db.pg` text rows, removing the per-request `json.encode` reflection on `/async-db`, `/crud` list and `/crud` GET.
- **crud cache is an id-indexed slab** (replaces `map[int]string`) with in-place buffer reuse + cache-aside invalidation, shared across ring workers under `RwMutex`.
- **`crud_list` uses a single windowed query** (`count(*) OVER()`) instead of a page `SELECT` + a separate `count(*)`.
- **`parse_crud_body_fast`** + borrowed JSON field parsers (`json.decode` fallback kept for escaped bodies).
- **`parse_i64_slice` / `dechunk_into` / `parse_hex_slice`** — allocation-free query/body parsing.

**Static:** unlike the epoll twin, this entry does **not** set `static_assets.sendfile_min_bytes`. The io_uring backend has no sendfile path (no `core.enable_sendfile` / `queue_file` drain), so a low threshold makes `static_assets` read every large `.br`/`.gz` sibling from disk per request (blocking the ring). Keeping the default (256 KiB) preloads all arena siblings (< 256 KiB) and serves them as a zero-copy `core.queue_buf` borrowed send.

## What is intentionally NOT changed

DB profiles (`fortunes`, `async-db`, `api-*`, `crud`) stay capped by the **blocking `db.pg`** on the single ring worker — the io_uring backend has no async runtime to await DB readiness on the ring (enghitalo/vanilla#83). A per-worker reused render scratch (dropping the small per-request DB-render buffers) is a follow-up now that io_uring supports `make_state` (enghitalo/vanilla#93 landed; entry follow-up enghitalo/vanilla#97).

## Validation

- Both images build (`v -prod -d vanilla_tls`).
- Ran **both** containers against a **pristine seeded Postgres** (fresh DB per framework) and diffed every route — `pipeline`, `baseline11` (positive **and negative**), `upload`, `json`, `json-comp`, `async-db`, `fortunes`, `static` (br negotiation), `crud` list, `crud` GET (MISS→HIT), `crud` create/update, 404, `json-tls` — **all 17 byte-for-byte identical to vanilla-epoll**.
- X-Cache verified: `GET` MISS → HIT, re-MISS after a `PUT` (slab invalidation); `POST` → 201; `json-comp` → `Content-Encoding: gzip`; `json-tls` → 200 over TLS 1.3.
- Static (the path that regressed in #965 before the fix): `wrk` at 64 conns on `/static/vendor.js` (→ 67 KB `.br`) serves ~**101k req/s / 6.38 GB/s, zero socket errors** (preloaded, bandwidth-bound).


enghitalo:
Pushed a follow-up commit (44ca0b6) that fully converges the io_uring entry onto the epoll twin, now that the io_uring backend supports per-worker state (enghitalo/vanilla#94, merged):

- Split `Shared` → `SharedRO` (process-shared data + mutex-guarded crud slab / gz cache) + per-worker `WorkerCtx { ro, scratch }`, dispatched via **`stateful_handler` + `make_state`** — same shape as `vanilla-epoll`.
- DB render paths (`write_async_db` / `write_crud_list` / `write_crud_get` MISS / `write_fortunes`) now render into the worker's **reused scratch** instead of a per-request `[]u8` — removes the api-* memory growth from the earlier run. High-RPS non-DB paths unchanged.
- Bumped the pinned vanilla lib `b189036` → `6fb4244` (includes #94; the old pin rejects `stateful_handler` on io_uring at `new_server()` and panics on boot).

Closes enghitalo/vanilla#97. The only remaining handler-code divergence from the epoll twin is blocking `db.pg` vs async `pg_async` (enghitalo/vanilla#83).

Re-verified: all 17 routes byte-identical to `vanilla-epoll` on a pristine seeded Postgres; wrk healthy on every path (pipeline 241k, baseline 242k, json 194k, static/vendor.js 98k, crud 237k rps, zero socket errors).

MDA2AV (approved):

PR #973, opened 2026-07-09
feat(engines): Three engines to find platform baseline for .NET / Java / Kotlin

Kaliumhexacyanoferrat:
Do not merge, this is only for data gaining purposes :)


Kaliumhexacyanoferrat:
PR is subject to be extended to make the technological ceilings visible to any user. Needs some work before we can merge that though.

PR #979, opened 2026-07-09
[WTX] Update HTTP/2

c410-f3r:
## Description

Ordinary update


MDA2AV:
You have a 725k loc log file, can you check that? Is that expected?

c410-f3r:
725k logs. Wow... Just... Wow...

Thanks for the heads up. The issue should be issued now.

MDA2AV (approved):

PR #980, opened 2026-07-09
[WTX] Update gRPC

c410-f3r:
## Description

Ordinary update


c410-f3r:
Zero clue why CI is falling. A local instance is normally returning a response when poked with `curl` so I need to investigate more.

PR #983, opened 2026-07-10
Test different heap size

ruroru:


ruroru:
Parallel gc
/benchmark -f ring-http-exchange

PR #984, opened 2026-07-10
[WTX] Add support for `baseline-h2`

c410-f3r:
## Description

Requires a different application because the port is different and the server demands certificates


c410-f3r:
If you don't mind, I would like to test the performance impact of aws-lc-rs, ring, openssl and graviola. Starting with the graviola backend.

c410-f3r:
https://github.com/MDA2AV/HttpArena/pull/979#issuecomment-4939777347

HTTP/2 over TLS is very similar to unencrypted h2c

c410-f3r:
Testing the performance of the `ring` crypto library.

/benchmark -f wtx-http2-tls

c410-f3r:
Looks like CI is taking a long time to finish jobs so I won't bother anymore. Instead of also testing `openssl` and `awc-ls-rs` the final cryptography backend will be `ring`. Thanks for the compute time.

/benchmark -f wtx-http2-tls --save

MDA2AV:
> Looks like CI is taking a long time to finish jobs so I won't bother anymore. Instead of also testing `openssl` and `awc-ls-rs` the final cryptography backend will be `ring`. Thanks for the compute time.

I need to approve the bench runs and have been a bit off lately but just add them, they will be accepted

MDA2AV (approved):

c410-f3r:
Thanks @MDA2AV !

PR #986, opened 2026-07-11
Add ktor-kormium (JVM) and ktor-kormium-native (Kotlin/Native) frameworks

knyazevs:
Adds two Kotlin entries that share one endpoint surface and differ only in runtime, so they can be compared directly:

## frameworks/ktor-kormium — JVM
- Ktor 3.5 on Netty, JDK 21 (same base/flags as the existing `ktor` entry)
- [Kormium ORM](https://github.com/kormium/kormium) 0.9.1 over PostgreSQL JDBC + HikariCP, suspend API
- Image: ~513 MB

## frameworks/ktor-kormium-native — Kotlin/Native
- The same Ktor app compiled to a self-contained linuxX64 binary (no JVM), CIO engine
- Kormium's native PostgreSQL driver (libpq via cinterop)
- Image: ~140 MB (debian-slim + libpq5)

Both subscribe to `baseline, limited-conn, pipelined, json, async-db, crud, fortunes`. TLS/H2/WS/static are deliberately left out: Ktor's native CIO engine has no TLS/HTTP2, and keeping the two test sets identical is the point of the pair.

Notes:
- The routing code mirrors the existing `ktor` (Exposed) entry almost 1:1 — only the database layer differs, so `ktor` vs `ktor-kormium` also reads as an Exposed-vs-Kormium comparison on identical plumbing.
- The native Dockerfile pins `FROM --platform=linux/amd64`: the output is a Kotlin/Native linuxX64 executable and must link against an amd64 libpq regardless of the build host arch (it produces a lint warning; happy to change if you prefer another approach).
- `scripts/validate.sh` passes 32/32 for both entries locally.

@BennyFranciscus — implementation questions welcome. Kormium maintainer here (see `maintainers` in both meta.json), happy to iterate.


knyazevs:
Closing for now — will reopen once we've refreshed the entries (dependency bumps + another local validation pass).

PR #988, opened 2026-07-11
Update simplew

stratdev3:
I want to check performance of the alpha before releasing


stratdev3:
fixed memory leak in websocket

/benchmark -f simplew

PR #993, opened 2026-07-12
[WTX] Add support for Unary TLS

c410-f3r:
## Description

Let's see if this case will hang like in https://github.com/MDA2AV/HttpArena/pull/980.


c410-f3r:
Yeap. I will investigate more about what is going on

c410-f3r:
Tried different approaches over several hours without success.

PR #994, opened 2026-07-12
[rails] Let Symbol#to_s return a frozen string

p8:
This should reduce string allocations.
Large codebases like Shopify and GitHub run this in production: https://bugs.ruby-lang.org/issues/22137
This is how Ruby 4.1 works:
https://github.com/ruby/ruby/pull/17781


p8:
Revert changes to return symbols for database calls:
/benchmark -f rails

MDA2AV (approved):

PR #997, opened 2026-07-13
zix ws entry

prothegee:
## Description

zix entry for ws using 0.5.x branch.

Update:
- Adding tweak buffer.

Checklist:
- [x] `/benchmark -f zix-ws`
- [x] `/benchmark -f zix-ws --save`

PR #982, opened 2026-07-10
zix grpc entry

prothegee:
## Description

zix entry for grpc using 0.5.x branch.

Update:
- Using dual listener port.

Checklist:
- [x] `/benchmark -f zix-grpc`
- [x] `/benchmark -f zix-grpc --save`

PR #998, opened 2026-07-13
zix http1 entry

prothegee:
## Description

zix entry for http1 using 0.5.x branch.

Update:
- Using dual listener port.
- Tweak on config init side.
- Adding new tests:
    - static-tls
    - json-tls
    - json-comp
    - api-4
    - api-16
    - async-db
    - crud

Checklist:
- [x] `/benchmark -f zix`
- [ ] `/benchmark -f zix --save`


prothegee:
Testing new approach and database tests entries for H1.

prothegee:
Oh my, didn't expect the cpu burn that much with high memory consumption.

prothegee:
Attempt 2 with 0.5.x-rc2 branch.

prothegee:
Submitting save result.

prothegee:
Testing async-db using new shards.

MDA2AV:
## Review: zix H/1.1 handlers — several tests replay pre-computed responses or skip the measured work

Hi @prothegee, thanks for the substantial submission, the engine itself is clearly fast. I went through the H/1.1 handlers against the profile specs and a few of them short-circuit the exact work each profile is designed to measure, which is inflating the results. File:line detail below. **`baseline`, `pipelined`, `limited-conn` and `async-db` all look correctly implemented**. The issues are concentrated in the JSON, upload, and crud handlers.

The following is a summed up analyzis by Claude Fable 5

### 1. `json` / `json-comp` / `json-tls` — full HTTP response cache

`src/handlers/json.zig` memoizes the fully-serialized HTTP response (headers + body, and the gzip variant) keyed by the request, then replays it verbatim:

```zig
// json.zig:52
if (want_gzip) {
    if (zix.Http1.cacheLookupEncoded(head, "gzip")) |cached| {
        zix.Http1.writeAllFD(fd, cached) catch {};   // replay pre-built bytes
        return;
    }
} else {
    if (zix.Http1.cacheLookup(head)) |cached| { zix.Http1.writeAllFD(fd, cached); return; }
}
```

The store side (`sendWithCacheFD` / `sendGzipCachedFD`) uses a 6-minute TTL:

```zig
// src/shared/cache.zig:7
pub const TTL_MS: u32 = (60 * 1_000) * 6;   // comment: "each key is built once per run"
```

The `json` profile only rotates 7 `(count, m)` pairs (`json-comp` 3), so every distinct response is serialized once in the first few milliseconds and then served as a raw `writeAllFD` memcpy for the rest of the run. These profiles end up measuring a socket write of a pre-built buffer, not the per-request JSON serialization (and gzip) they're defined to measure. This is what the `json-processing` spec rules out ("no pre-serialized / response cache") and what the Tuned rules list under *not allowed* — "pre-computed response bodies" and "response caching: memoizing the full HTTP response body keyed by URL/params and replaying it". (It's distinct from allowed upstream *data* caching like DB rows/JWT — here it's the rendered HTTP response itself.)

The results match: zix's `json-comp` (3.13M rps, #1) comes out **faster than its own plain `json`** (2.36M) — compression should cost more, not less, unless the compressed bytes are pre-built and replayed. And since `json-comp` scores `rps × (minBpr/myBpr)²`, a cached gzip body wins on both raw rps and the compression-ratio bonus. The same `/json` handler backs the JSON portion of `api-4`/`api-16`, so those inherit it too.

**Fix:** serialize (and compress) per request; remove the response cache from the `/json` path.

### 2. `upload` — returns Content-Length instead of the bytes read

```zig
// src/handlers/upload.zig:22
const n: u64 = if (head.content_length > 0) head.content_length else body.len;
```

This returns the `Content-Length` header value and drains the body rather than ingesting it (the file comment says as much: *"Content-Length is authoritative, the engine drains oversized bodies"*). The `upload` spec requires reading the body and returning the actual received byte count — **returning the Content-Length value is explicitly disallowed** — precisely because echoing the header lets a server skip the 20 MB ingest. The numbers reflect that: zix is #1 at 6,726 rps @256c and 8,232 @32c, ~2.1–2.6× the next engine (ioxide 3,179 / typev 3,152). A 20 MB body can't be ingested per request 2.5× faster than every other io_uring/epoll server — it isn't being read.

**Fix:** count the bytes actually read off the socket and return that count.

### 3. `crud` — single-item cache TTL is 5 s, spec is 200 ms

```zig
// src/shared/crudcache.zig:26
const ITEM_TTL_MS: i64 = 5000;   // spec: 200 ms
```

`GET /crud/items/{id}` is defined with a **200 ms** absolute TTL. At 5,000 ms the cached body lives 25× longer; with reads randomized across 50k IDs and only 15% writes (which invalidate only the id they touch), most items stay hot for their full 5 s, so the hit rate is much higher than the profile intends and the bulk of reads never reach Postgres. The 200 ms bound is what keeps the cache-aside path under real DB pressure — which is the thing being measured.

Relatedly, list responses are cached 1 s with stale-serving (`LIST_TTL_MS = 1000`, crudcache.zig:28) even though the list endpoint is specified as a live `SELECT … LIMIT … OFFSET` with no cache.

**Fix:** set the single-item read TTL to 200 ms; drop the list-response cache (or confirm the intended list semantics).

### Correctly implemented 👍
- `baseline` / `pipelined` / `limited-conn` — computed per request; fixed body where the profile specifies one.
- `async-db` — genuinely queues a Postgres query per request, no cache.

### Summary
The JSON, upload, and crud handlers each bypass the specific work their profile measures (serialization, 20 MB ingest, DB-pressured cache-aside). Could you rework those three before a `--save` run? Happy to clarify any spec detail if something reads as ambiguous — the rest of the entry looks great where it's doing the real work.

prothegee:
> ## Review: zix H/1.1 handlers — several tests replay pre-computed responses or skip the measured work
> 
> Hi @prothegee, thanks for the substantial submission, the engine itself is clearly fast. I went through the H/1.1 handlers against the profile specs and a few of them short-circuit the exact work each profile is designed to measure, which is inflating the results. File:line detail below. **`baseline`, `pipelined`, `limited-conn` and `async-db` all look correctly implemented**. The issues are concentrated in the JSON, upload, and crud handlers.
> 
> The following is a summed up analyzis by Claude Fable 5
> ### 1. `json` / `json-comp` / `json-tls` — full HTTP response cache
> 
> `src/handlers/json.zig` memoizes the fully-serialized HTTP response (headers + body, and the gzip variant) keyed by the request, then replays it verbatim:
> 
> ```zig
> // json.zig:52
> if (want_gzip) {
>     if (zix.Http1.cacheLookupEncoded(head, "gzip")) |cached| {
>         zix.Http1.writeAllFD(fd, cached) catch {};   // replay pre-built bytes
>         return;
>     }
> } else {
>     if (zix.Http1.cacheLookup(head)) |cached| { zix.Http1.writeAllFD(fd, cached); return; }
> }
> ```
> 
> The store side (`sendWithCacheFD` / `sendGzipCachedFD`) uses a 6-minute TTL:
> 
> ```zig
> // src/shared/cache.zig:7
> pub const TTL_MS: u32 = (60 * 1_000) * 6;   // comment: "each key is built once per run"
> ```
> 
> The `json` profile only rotates 7 `(count, m)` pairs (`json-comp` 3), so every distinct response is serialized once in the first few milliseconds and then served as a raw `writeAllFD` memcpy for the rest of the run. These profiles end up measuring a socket write of a pre-built buffer, not the per-request JSON serialization (and gzip) they're defined to measure. This is what the `json-processing` spec rules out ("no pre-serialized / response cache") and what the Tuned rules list under _not allowed_ — "pre-computed response bodies" and "response caching: memoizing the full HTTP response body keyed by URL/params and replaying it". (It's distinct from allowed upstream _data_ caching like DB rows/JWT — here it's the rendered HTTP response itself.)
> 
> The results match: zix's `json-comp` (3.13M rps, #1) comes out **faster than its own plain `json`** (2.36M) — compression should cost more, not less, unless the compressed bytes are pre-built and replayed. And since `json-comp` scores `rps × (minBpr/myBpr)²`, a cached gzip body wins on both raw rps and the compression-ratio bonus. The same `/json` handler backs the JSON portion of `api-4`/`api-16`, so those inherit it too.
> 
> **Fix:** serialize (and compress) per request; remove the response cache from the `/json` path.
> ### 2. `upload` — returns Content-Length instead of the bytes read
> 
> ```zig
> // src/handlers/upload.zig:22
> const n: u64 = if (head.content_length > 0) head.content_length else body.len;
> ```
> 
> This returns the `Content-Length` header value and drains the body rather than ingesting it (the file comment says as much: _"Content-Length is authoritative, the engine drains oversized bodies"_). The `upload` spec requires reading the body and returning the actual received byte count — **returning the Content-Length value is explicitly disallowed** — precisely because echoing the header lets a server skip the 20 MB ingest. The numbers reflect that: zix is #1 at 6,726 rps @256c and 8,232 @32c, ~2.1–2.6× the next engine (ioxide 3,179 / typev 3,152). A 20 MB body can't be ingested per request 2.5× faster than every other io_uring/epoll server — it isn't being read.
> 
> **Fix:** count the bytes actually read off the socket and return that count.
> ### 3. `crud` — single-item cache TTL is 5 s, spec is 200 ms
> 
> ```zig
> // src/shared/crudcache.zig:26
> const ITEM_TTL_MS: i64 = 5000;   // spec: 200 ms
> ```
> 
> `GET /crud/items/{id}` is defined with a **200 ms** absolute TTL. At 5,000 ms the cached body lives 25× longer; with reads randomized across 50k IDs and only 15% writes (which invalidate only the id they touch), most items stay hot for their full 5 s, so the hit rate is much higher than the profile intends and the bulk of reads never reach Postgres. The 200 ms bound is what keeps the cache-aside path under real DB pressure — which is the thing being measured.
> 
> Relatedly, list responses are cached 1 s with stale-serving (`LIST_TTL_MS = 1000`, crudcache.zig:28) even though the list endpoint is specified as a live `SELECT … LIMIT … OFFSET` with no cache.
> 
> **Fix:** set the single-item read TTL to 200 ms; drop the list-response cache (or confirm the intended list semantics).
> ### Correctly implemented 👍
> 
>     * `baseline` / `pipelined` / `limited-conn` — computed per request; fixed body where the profile specifies one.
> 
>     * `async-db` — genuinely queues a Postgres query per request, no cache.
> 
> 
> ### Summary
> 
> The JSON, upload, and crud handlers each bypass the specific work their profile measures (serialization, 20 MB ingest, DB-pressured cache-aside). Could you rework those three before a `--save` run? Happy to clarify any spec detail if something reads as ambiguous — the rest of the entry looks great where it's doing the real work.

Great to know, working on it.

prothegee:
Fixed has been made.
Testing normal bench before assign `--save` run.

prothegee:
Below the adjustment/reworked as follows:

__*json / json-comp / json-tls:*__

Requested: serialize (and compress) per request, remove the response cache from the `/json` path.

Change (`frameworks/zix/src/handlers/json.zig`, `frameworks/zix/src/main.zig`):

- The replay path is gone: no response lookup, no stored responses, no TTL.
- Every request renders the JSON body from the loaded dataset, assembles the
  header, and writes it out per request.
- gzip responses compress per request from the freshly rendered body.
- The engine response-cache configuration is removed from `main.zig` (nothing
  uses it anymore), so there is no response cache anywhere on the entry.

<br>

__*upload:*__

Requested: count the bytes actually read off the socket and return that count.

Change (`frameworks/zix/src/handlers/upload.zig` plus the vendored zix engine):

- The engine's receive path now counts every byte it consumes off the socket
  for a body larger than the receive buffer, completes the body before the
  handler runs, and exposes the counted total as `Request.bodyReceived()`.
- The handler returns that counted total. The `Content-Length` header value is
  no longer used for the response.

<br>

__*crud:*__

Requested: single-item read TTL 200 ms, drop the list-response cache.

Change (`frameworks/zix/src/shared/crudcache.zig`, `frameworks/zix/src/handlers/crud.zig`, `frameworks/zix/src/shared/dbpg.zig`):

- Single-item TTL is 200 ms absolute (writes still invalidate the id on PUT,
  cache-aside with `X-Cache: MISS` / `X-Cache: HIT` unchanged).
- The list-page cache is removed entirely (slots, lookups, stores, and both
  call sites), so `GET /crud/items` executes the live
  `SELECT ... LIMIT ... OFFSET` on every request.

prothegee:
> Below the adjustment/reworked as follows:
> 
> **_json / json-comp / json-tls:_**
> 
> Requested: serialize (and compress) per request, remove the response cache from the `/json` path.
> 
> Change (`frameworks/zix/src/handlers/json.zig`, `frameworks/zix/src/main.zig`):
> 
>     * The replay path is gone: no response lookup, no stored responses, no TTL.
> 
>     * Every request renders the JSON body from the loaded dataset, assembles the
>       header, and writes it out per request.
> 
>     * gzip responses compress per request from the freshly rendered body.
> 
>     * The engine response-cache configuration is removed from `main.zig` (nothing
>       uses it anymore), so there is no response cache anywhere on the entry.
> 
> 
> 
> **_upload:_**
> 
> Requested: count the bytes actually read off the socket and return that count.
> 
> Change (`frameworks/zix/src/handlers/upload.zig` plus the vendored zix engine):
> 
>     * The engine's receive path now counts every byte it consumes off the socket
>       for a body larger than the receive buffer, completes the body before the
>       handler runs, and exposes the counted total as `Request.bodyReceived()`.
> 
>     * The handler returns that counted total. The `Content-Length` header value is
>       no longer used for the response.
> 
> 
> 
> **_crud:_**
> 
> Requested: single-item read TTL 200 ms, drop the list-response cache.
> 
> Change (`frameworks/zix/src/shared/crudcache.zig`, `frameworks/zix/src/handlers/crud.zig`, `frameworks/zix/src/shared/dbpg.zig`):
> 
>     * Single-item TTL is 200 ms absolute (writes still invalidate the id on PUT,
>       cache-aside with `X-Cache: MISS` / `X-Cache: HIT` unchanged).
> 
>     * The list-page cache is removed entirely (slots, lookups, stores, and both
>       call sites), so `GET /crud/items` executes the live
>       `SELECT ... LIMIT ... OFFSET` on every request.

Requesting review before `--save` and `merge` @MDA2AV

prothegee:
Ready to review before merge, @MDA2AV

MDA2AV:
> Ready to review before merge, @MDA2AV

Beem kinda busy, I'll do a full review tommorow, need some time to do it properly

prothegee:
> Beem kinda busy, I'll do a full review tommorow, need some time to do it properly

Sure. Hope you doin well.

Regards.

MDA2AV:
Ok, I reviewed again. Some things improved but some issues remain. 

On the upload test, I must say your AI agent really went all in on this test.. technically following the rules but doing something no server would ever do. You count the bytes received but discard them and never read into zix's memory.. it is expected that a server reads the request data to its memory. As this is an engine and test is not scored, we can allow this but keep in mind this would never work on a production server.

On json tests, you need to serialize per request, not caching the serialized data at startup and then serve it. This one is important as json-tls is a scored test.

There are a lot other things zix is doing that are clearly benchmark oriented and would be a very big problem in a production use, however this is engine section so I won't pick on that but if you are planning on using it as an engine for a real server, there are quite a lot of issues.

On CRUD, the ttl seems to be addressed.

Regardless, it is an interesting project to push benchmark numbers! :)

prothegee:
Yes, zix.Http1 intent really rely on user implementation and promote for user control.
But since you brought that up, now I can add more tests to cover.

I knew it's engine category, but engine correctness is considerable aspect (despite some zix engine/s features).
So if there's a rule, it follow, but the engine intent should be adjusted again.

Converting to draft.
I'll mention you again after the fix, @MDA2AV 

Thank you.

prothegee:
Bump zix branch to 0.5.x

and follow-up https://github.com/MDA2AV/HttpArena/pull/1121#issuecomment-5295835172

prothegee:
Hi, open to review for this pull-request.
@MDA2AV @Kaliumhexacyanoferrat

prothegee:
https://github.com/MDA2AV/HttpArena/pull/997 and https://github.com/MDA2AV/HttpArena/pull/982 already merged, but this p.r. skipped.

https://github.com/MDA2AV/HttpArena/pull/998/changes/83aa9f0f0e25e53042b940e722dc641dfbf55304 already done.

Looking forward.
@Kaliumhexacyanoferrat @MDA2AV

MDA2AV:
> Ok, I reviewed again. Some things improved but some issues remain.
> 
> On the upload test, I must say your AI agent really went all in on this test.. technically following the rules but doing something no server would ever do. You count the bytes received but discard them and never read into zix's memory.. it is expected that a server reads the request data to its memory. As this is an engine and test is not scored, we can allow this but keep in mind this would never work on a production server.
> 
> On json tests, you need to serialize per request, not caching the serialized data at startup and then serve it. This one is important as json-tls is a scored test.
> 
> There are a lot other things zix is doing that are clearly benchmark oriented and would be a very big problem in a production use, however this is engine section so I won't pick on that but if you are planning on using it as an engine for a real server, there are quite a lot of issues.
> 
> On CRUD, the ttl seems to be addressed.
> 
> Regardless, it is an interesting project to push benchmark numbers! :)

Did you look into these points?

prothegee:
> Did you look into these points?

Yes, crud ttl already addressed.

---

For the upload you can verify it in the [handler implementation](https://github.com/prothegee/HttpArena/blob/zix-http1/frameworks/zix/src/handlers/upload.zig).

It physically reads and processes the data into memory.

First it read into engine memory,
the handler calls `const body = try req.body();` this made the engine handler read the payload off the socket and returns a memory slice containing the actual uploaded data.

Then the handler explicitly copies those bytes into local buffer.
```zig
const taken = @min(body.len, SINK_BYTES);
@memcpy(tl_sink[0..taken], body[0..taken]);
```

Finally, the output is checked from `received` from `req.bodyReceived();`,
copied to local `body_buf` and send it as the correct size as `out`.
```zig
const received: u64 = req.bodyReceived();

var body_buf: [COUNT_BUF]u8 = undefined;
const out = try std.fmt.bufPrint(&body_buf, "{d}", .{received});

try zix.Http1.sendSimpleFD(fd, @intFromEnum(zix.Http1.Status.Code.OK), zix.Http1.Content.Type.TEXT_PLAIN.asString(), out);
```

[Reference test of the size to be expected equal](https://github.com/prothegee/zix/blob/main/src/tcp/http1/request.zig#L354).


---

Now for the json handler.
You can verify it in the [handler implementation](https://github.com/prothegee/HttpArena/blob/zix-http1/frameworks/zix/src/handlers/json.zig) and [dataset loader](https://github.com/prothegee/HttpArena/blob/zix-http1/frameworks/zix/src/shared/dataset.zig).

It physically serializes the json payload per request, not caching the serialized string at startup (no initialize data cache in main).

First it loads the dataset into engine memory as raw typed values (integers, strings, structs), not as pre-rendered json bytes.
```zig
// dataset.zig
const items = try jzon.deserialize([]const Item, allocator, raw, .{
    .strategy = .GENERATED,
    .strings = .BORROW,
});
```

Then the handler dynamically reads the request query parameter `m` (multiplier). Because the `total` field depends on this per-request value, the json body cannot be statically cached and must be computed on the fly.
```zig
// json.zig
const multiplier: u64 = if (zix.Http1.queryParam(head, "m")) |raw|
    std.fmt.parseInt(u64, raw, 10) catch 1
else
    1;
```

Finally, the handler serializes the json body on every single request into a local buffer (or directly into the engine's send buffer), and sends it.
```zig
// json.zig
const body_len = renderBody(body, rows, count, multiplier) catch {
    try response.badRequest(fd);
    return;
};

// ... then sends it dynamically
```

When the processing data fails, it will call `try response.badRequest(fd);` in the process.

[Reference of the dynamic rendering](https://github.com/prothegee/HttpArena/blob/zix-http1/frameworks/zix/src/handlers/json.zig#L104).

MDA2AV:
Ok, great! I will take a look at it again, this entry takes some time to review because of the size of it.
Sorry that this is taking a bit longer, thank you for your patience!

prothegee:
Sure. I'll keep updating from the main branch too.

MDA2AV:
Ok, I've reviewed again and there is modification required and a few points to say.

So, for json tests, you need to use an actual serialization library.. something that is production ready, otherwise what are we even measuring here, anyone can rollout a lean json serializer and makes these results pointless.

Other thing is I have to ask if this project is just a benchmark tool or actually plans on going production one day, I've looked at a few things and for example writing your own tls 1.3 layer and RSA signing is just a terrible idea, also it doesn't cover a fraction of real world usage and is full of vulnerabilities, consider using an actual library like OpenSSL.

prothegee:
Pushing new commit to address static serve https://github.com/MDA2AV/HttpArena/pull/1209

Need new benchmark run.

---

Next: follow-up the maintainers previous comment.

prothegee:
Hi, thank you for the review and feedback, @MDA2AV 

To answer your question regarding the project's intent:
Zix is already used in my production work for gRPC and HTTP/1 since 0.2.0 and all decision records in ADRs with HLDs & LLDs respectively.
Its core design philosophy stood with *push Zig std/zero-dependencies*. This context might help clarify the design choices you pointed out.


__*1. Regarding the JSON Subject*__
Initially, I considered using existing serialization libraries.
However, many available Zig JSON libraries were either outdated, archived for the specific compiler version I target,
or introduced unwanted dependency overhead. Since Zix strictly adheres to a zero-dependency philosophy, I decided to build a custom tooling suite (`jzon`).

The goal here is to measure the raw performance of the HTTP engine without being bottlenecked or skewed by a third-party JSON library's implementation.
This custom serializer is tested against RFCs, edge cases, and behavior specifications.


__*2. Regarding Production Readiness & Custom TLS*__
I agree with you that rolling your own TLS/crypto is a complex idea for a general-purpose web server exposed to the public internet.

Because Zix's philosophy mandates zero external dependencies (no C-FFI, no OpenSSL), a custom TLS 1.2/1.3 implementation was necessary to complete the stack.
This custom TLS layer is not intended to replace established libraries like OpenSSL for public-facing applications,
but rather to serve as an isolated, pushing the Zig standard library (before split it into another stand-alone projects),
zero-deps transport layer for private/internal networks (where it is currently used). It has been strictly validated against RFCs through unit, integration, and behavior tests.


I can provide you with what's been tested both for the json & tls.

My primary goal for submitting to the engine category was to measure the production isolated test,
raw overhead of the HTTP engine itself, free from external C-FFI or dependency bottlenecks for the minimum dependency isolation.

If the rules for the engine category strictly mandate the use of established external libraries to be considered a valid production benchmark, please let me know.
I want to ensure my submission aligns with the community's expectations and the benchmark's goals.

Best regards,
Prana Ron

MDA2AV:
As an engine submission, these points are allowed because there are no strict rules, the reason why I am pushing on some points is simply because if eventually you decide to add an emerging entry for example, a lot of points will collide with the rules for framework entries, namely using standard production level established libraries for serialization, tls, etc.

My core idea for engine entries was to include engines that frameworks use to measure the framework overhead, for example node vs a framework that uses node, it is very likely in a near future Ill need to restructure how it operates and either add a new tier or a separation between some engines.

For now I will accept this since you are not breaking a rule and actually serializing and encrypting, enough to pass validation tests. In the future I will also reinforce the validation suite to ensure more confidence in the entries we benchmark.

prothegee:
Hi, @MDA2AV 
Thank you for quick follow-up response.

```
As an engine submission, these points are allowed because there are no strict rules, the reason why I am pushing on some points is simply because if eventually you decide to add an emerging entry for example, a lot of points will collide with the rules for framework entries, namely using standard production level established libraries for serialization, tls, etc.
```
The `emerging` wasn't my intent at all.
Now I see one clear problem. So I think the result on the latest bench.
image
image

Those entries shouldn't exists. So I'll make follow-up pull-request later on.
The entry should be all as `zix` and just in `engine` category.

---

```
For now I will accept this since you are not breaking a rule and actually serializing and encrypting, enough to pass validation tests. In the future I will also reinforce the validation suite to ensure more confidence in the entries we benchmark.
```
That would be lovely, all entries can push their implementations.
And HttpArena help those by validating best benchmark check for the long run of their framework/library.

Looking forward!

MDA2AV (approved):

PR #1002, opened 2026-07-17
feat(aspnet-minimal): Move business logic and types into reusable units

Kaliumhexacyanoferrat:
Idea is that we can easily share (copy) the implementation with other C# frameworks.


Kaliumhexacyanoferrat:
Changes on H2/H3 are unrelated and either due to .NET updates or changes to the benchmarking infrastructure.

MDA2AV (approved):

PR #1006, opened 2026-07-18
feat(fasthttp): append-handler architecture + baseline/limited-conn/api-4/api-16 (supersedes #985)

enghitalo:
## What

Rebuilds the **fasthttp** entry on V's new fasthttp backend from [vlang/v#27771](https://github.com/vlang/v/pull/27771) — the *"refactor toward the vanilla architecture"* commit — and expands it to the full isolated-profile suite.

The handler now uses the **zero-copy append handler**: it writes the raw HTTP response straight into the connection's reused write buffer (no per-request response object), reads a lock-free per-worker scratch buffer via **`make_state`**, streams static bodies with **`sendfile(2)`** (`ResponseControl.file_path`), and relies on the backend's deterministic request **framing** to reassemble TCP-fragmented and chunked requests.

## Relationship to #985

This **supersedes** the classic-`!HttpResponse` full suite merged in #985. It is a strict superset: the same 8 tests **plus** `baseline`, `limited-conn`, `api-4`, `api-16` (12 total), on a faster zero-copy architecture.

| Test | Endpoint |
|------|----------|
| `baseline` / `limited-conn` | `GET/POST /baseline11?a&b` (handles chunked; reactor reassembles fragmented requests) |
| `pipelined` | `GET /pipeline` (precomputed constant) |
| `json` | `GET /json/{count}?m=M` (precomputed prefixes, single pass into the write buffer) |
| `json-comp` | `GET /json/{count}?m=M` + gzip (process-shared cache) |
| `upload` | `POST /upload` |
| `static` | `GET /static/*` via `sendfile(2)` from a preloaded MIME/size table |
| `async-db` | `GET /async-db?min&max&limit` |
| `crud` | `GET/POST /crud/items`, `GET/PUT /crud/items/{id}` (cache-aside, `X-Cache` MISS/HIT) |
| `fortunes` | `GET /fortunes` (DB rows + runtime row, HTML-escaped) |
| `api-4` / `api-16` | mixed `/baseline11` + `/json` + `/async-db` |

`json-tls` is **not** subscribed: the stdlib fasthttp server has no TLS backend. DB paths (`async-db`/`crud`/`fortunes`) issue blocking `db.pg` queries on the worker thread — fasthttp has no async watch reactor yet (`.suspend` drops the connection).

## Notes

- V pin bumped `cef604a26` → `84a76d791` (a superset-descendant of the old pin: keeps db.pg#27265, array-push#27470, Boehm GC #27560). No later master commit touches `vlib/fasthttp`, so the pin is exactly the module code validated.
- Migrated `json` → `json2`: the deprecated `json` module is a **hard compile error under `-prod`** on modern V master; `json2` is a drop-in with byte-identical encode output.
- Shared read-only data + the mutex-guarded crud/gzip caches live in `user_data`; per-worker render scratch comes from `make_state`. The crud cache is process-shared so the `X-Cache` MISS→HIT probe survives `SO_REUSEPORT` routing.

## Verification

Compiles with `-prod`. Every endpoint was checked end-to-end against a seeded Postgres — as a host binary **and** in the built container image (slim runtime): baseline sums (incl. chunked + TCP-fragmented reassembly), json schema + totals at count=50, json-comp encoding on/off + round-trip, upload byte count, static Content-Type/size/404, async-db count/schema/empty-range, crud list/get/cache-MISS→HIT/create/update-invalidation, fortunes doctype/`<tr>` count (202)/runtime row/XSS escape/size.


enghitalo:
@MDA2AV Can this PR be merged?

MDA2AV (approved):

PR #1029, opened 2026-07-24
Add a page search to the top of the left menu

### LINKED ISSUE #970: Re-add search functionality for pages
With the new design, the website search bar has been removed. When looking for a specific test specification, it was quite helpful.

MDA2AV:
Closes #970.

The Hextra theme's search went away with Hugo, and with it the only way to find a test specification by name. The board's existing search box filters *framework rows*, so a query like `prior-knowledge` or `MTU` matched nothing anywhere on the site — and 124 Knowledge Base pages were reachable only by walking the tree.

### What it searches

- **The Knowledge Base** — page titles **and full body text**, so `MTU` finds the setup page and `prior-knowledge` finds the H/2 validation docs.
- **Every leaderboard destination** — 28 profile views and the 6 composite scopes. `h2c` reaches both the *JSON (h2c)* board and its implementation page.

Results are ranked (title hit > body hit, exact title match wins outright, views float slightly above docs since they're the cheaper landing). Multi-word queries are AND. Each docs hit carries a breadcrumb and a snippet with the match highlighted. Capped at 25.

### No dependency, no index file

`docs.js` already ships every page's rendered HTML, so the index is built from what's already in memory — **158 entries over 0.24 MB of text in ~5 ms**, on the first keystroke rather than at load. Tags are stripped with a regex rather than by assigning to a detached `innerHTML`, which would fetch any `` the docs grow later.

### Two structural details

- **`buildNav()` now clears `#navTree`, not `#nav`.** The search box is static markup above it, so it survives the nav rebuild that every navigation triggers — otherwise typing would lose focus the moment results changed.
- **On mobile**, the tree is still hidden in favour of the existing `<select>`, but `.nav` is no longer `display:none`, so the search bar stays reachable.

Keyboard: arrows move the selection, Enter opens, Escape clears. Opening a result clears the query first, so you land on a normal menu showing where you now are. Colours come from the existing theme variables — light and dark both work.

### Verified end-to-end in headless Chrome against the real data files

| check | result |
|---|---|
| tree builds into its new container | 3 sections ✓ |
| `fortunes` → hits, tree hidden, first selected | 6 hits ✓ |
| ArrowDown moves selection | ✓ |
| click docs hit → page renders | "Fortunes (Templates)" ✓ |
| URL stays deep-linkable | `#doc=test-profiles/h1/isolated/fortunes` ✓ |
| input clears, tree returns | ✓ |
| nonsense query → empty state | ✓ |
| Escape restores | ✓ |
| 1400px and 400px layouts | ✓ (mobile keeps search, hides tree) |

Possible follow-up, deliberately left out to keep this scoped: a `/` or `Ctrl-K` shortcut to focus the box.

### DIFF
--- a/site/static/new-leaderboard/index.html
+++ b/site/static/new-leaderboard/index.html
@@ -95,6 +95,24 @@
   .nav::-webkit-scrollbar-track{background:transparent}
   .nav-mobile{display:none}
 
+  /* page search — pinned to the top of the menu so it stays put while the tree scrolls */
+  .nav-search{position:sticky; top:0; z-index:2; background:var(--bg); padding:.1rem 0 .55rem}
+  .ns-box{position:relative}
+  .ns-box svg{position:absolute; left:.6rem; top:50%; transform:translateY(-50%); color:var(--muted); pointer-events:none}
+  .nav-search input{font:inherit; font-size:.82rem; width:100%; padding:.45rem .6rem .45rem 1.85rem; background:var(--panel); color:var(--text); border:1px solid var(--line); transition:border-color .15s ease, box-shadow .15s ease}
+  .nav-search input:focus{outline:none; border-color:var(--accent); box-shadow:0 0 0 3px var(--accent-weak)}
+  .nav-search input::-webkit-search-cancel-button{cursor:pointer}
+  .nav-results{margin-bottom:.7rem}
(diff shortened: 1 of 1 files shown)


MDA2AV:
Closing as a duplicate — this is already in `main` via #1024 (squash-merged as `cfbd513d`). Same cause as #1028: I re-created it from the branch after the API outage cleared, without seeing that you'd already merged it.

PR #1035, opened 2026-07-24
test: --save pipeline check for the per-framework results layout

MDA2AV:
**Pipeline test for #1034 — please don't merge until we've looked at the result.**

Branched from `per-framework-results`, so its tree already has the per-framework results layout. The framework change itself is small and real (apt list cleanup + `--no-install-recommends` in bjoern's image layer), so the run has something legitimate to build.

### What we're checking

A `--save` run should update **only** `site/data/results/bjoern.json` — specifically its `baseline-512` and `baseline-4096` keys (currently `baseline-512` = 12,390 rps) — and touch no other framework's file. Under the old layout this would have rewritten `site/data/baseline-512.json` and `baseline-4096.json`, which every framework shares.

### How it has to be triggered

`/benchmark` comments run `pr-commands.yml` **from the default branch**, which then dispatches `benchmark-pr.yml` **also from the default branch**. So commenting here would run `main`'s commit step (staging the old flat paths) against this branch's scripts (writing per-framework files) — it would save nothing and prove nothing.

To exercise #1034's own workflow code, dispatch it explicitly against that branch:

```
gh workflow run benchmark-pr.yml --ref per-framework-results \
  -f pr=<this PR number> -f framework=bjoern -f profile=baseline -f save=true
```

It still requires your approval on the `runner` environment before anything executes.

Once #1034 is merged this stops being necessary — `benchmark-pr.yml` already merges `main` into the PR branch on `--save` runs, so ordinary `/benchmark` comments pick the new layout up automatically.


MDA2AV:
Closing — this was branched off `per-framework-results` before #1034 merged, so it carries every one of those changes instead of just the framework tweak. Superseded by a clean one-file branch off current `main`.

Issue #975, opened 2026-07-09
Show run logs when tests fail


MDA2AV:
When server doesn't start/come up logs are not accessible. 

This should improve troubleshooting/debug experience

PR #1055, opened 2026-07-26
[WTX] Add support for Unary TLS

c410-f3r:
## Description

Reopening #993 now that #975 is fixed to see if it is possible to figure out what is going on


c410-f3r:
Looks like validation tests don't map the "/certs" directory when running the image. Let's see.

c410-f3r:
```bash
grpc_needs_tls=false
if has_test "unary-grpc-tls" || has_test "stream-grpc-tls"; then
    grpc_needs_tls=true
fi

```

Even after injecting TLS certificates the validation test still doesn't work -> https://github.com/MDA2AV/HttpArena/actions/runs/30222679120/job/89847685310

Well, I guess I will just stay with the actual benchmarks.

MDA2AV (approved):

c410-f3r:
Logs were used to help with the CI issue. Unfortunately, I forgot to remove them.

It is still possible to historically ignore the `wtx-grpc-tls.log` files using `git rm`.

PR #1056, opened 2026-07-26
[WTX] Update gRPC

c410-f3r:
## Description

Reopens #980.

Looks like all validation tests for gRPC are not working


c410-f3r:
Erhhh... Wrong directory name...

/benchmark -f wtx-grpc --save

PR #1057, opened 2026-07-27
hical: pin HICAL_REF to 2a45016 (latest main) + trigger re-benchmark

Hical61:
## Description

Pin `HICAL_REF` from `main` to commit `2a45016`, ensuring reproducible builds.

Since the initial PR (#864, Jun 13), Hical's main branch has accumulated **41 commits**, including performance-relevant changes:
- `6062f54` perf: optimistic sync write + early readBuf release (throughput improvement under high concurrency)
- `633d181` perf: speculative read on stack buffer, eliminating ~20k unnecessary epoll_ctl calls
- `0894b2b` perf: idempotent preparePayload + WebSocket frame buffer reuse (P99 latency improvement)
- `7416731` perf: ChunkedBody/SSE/Gzip compression optimizations
- `06a4768` feat: PerfectHashRouter — compile-time perfect hash for static route lookup
- `071ce4c` feat: GzipCompression middleware (v2.6.6, enables json-comp)

## Changes

- `frameworks/hical/Dockerfile`: Switch from `git clone --branch main` to `git init + fetch + checkout FETCH_HEAD`, pinning `HICAL_REF` to `2a45016bbdcbf3dd93838beaf058e109f3f5dbfd`

Trigger re-benchmark to reflect the latest code on the leaderboard.

---

/benchmark -f hical --save


Hical61:
CI benchmark results look uneven — the static file test shows a major regression (RPS dropped from ~120K to ~22K, -81%), likely caused by one of the 41 commits between the pinned refs. We'll investigate the root cause, fix it, and trigger another benchmark run after that.

Hical61:
Fixed the static file regression. Pinned HICAL_REF to be8bf5d — ready for another run.

/benchmark -f hical --save

Hical61:
Scoped Gzip middleware to /json route group only — static file serving no longer goes through deflate. Ready for another run.

/benchmark -f hical --save

Hical61:
Fixed missing  —  only has a forward declaration, the full definition is needed for . Ready for another run.

/benchmark -f hical --save

Hical61:
Gzip middleware is now scoped to the `/json` route group instead of being global — static file serving is back to normal, and baseline/pipelined/limited-conn also benefited since they no longer pass through the gzip middleware chain. json-comp saw a minor dip from the extra coroutine frame in route-group middleware, within noise. Ready to merge if this looks good.

Thanks for the patience running the benchmarks and helping debug the regression — much appreciated!

Kaliumhexacyanoferrat:
> Thanks for the patience running the benchmarks and helping debug the regression — much appreciated!

That is what we want to provide, no four-week-cycles as we had with TechEmpower - so don't hesitate to request benchmarks.

> static file serving is back to normal

Short clarifying question, the benchmark still shows -50% for static/6800. Are you sure we should merge this?

Hical61:
Two more fixes landed:
1. PathCache LRU promotion now uses sampling (1/256 hits) instead of grabbing `unique_lock` on every cache hit — eliminates the lock contention that caused -38%/-48% regression at 4096/6800 connections.
2. Removed `isTextMime` branching — all files now go through `setFileBody` streaming send (64KB chunks, zero heap alloc). The text-file `setBody` path was left over from when gzip was global.

Ready for another run.

/benchmark -f hical --save

Hical61:
HICAL_REF pinned to bde1320 (restore isTextMime branching + lock-free thread_local PathCache).

Changes:
1. **Restore `isTextMime()`** — text files go through `setBody` (2 coroutine suspensions, single scatter-gather write) instead of `setFileBody` (3 suspensions + sync `open()` blocking io_context)
2. **Lock-free `thread_local` PathCache** — eliminates cross-core cache-line bouncing from the old global `shared_mutex`, expected to fix the -38%/-48% regression at 4096/6800 connections

Ready for another run.

/benchmark -f hical --save

Hical61:
Fixed the commit hash typo in Dockerfile. Ready for another run.

/benchmark -f hical --save

Hical61:
Thank you very much, I understand now. Static/4096 dropped 39%, static/6800 dropped 49%, and it has nothing to do with locks - thread-local PathCache was never locked. The problem lies in the fact that each request must open, read, and close the same file, with 64 threads running 4096 concurrent runs each, and file I/O directly filling up the kernel file system layer. The CPU is only 938%/813%, and 6400% of the quota has not been touched at all. All threads are stuck waiting for I/O.

Kaliumhexacyanoferrat:
Just a quick note - there are frameworks saturating CPU on static, serving roughly 1 mio requests/second, so there is surely no I/O induced bottleneck there at this level. Typically the kernel will map the file to memory as it is read more often so there is no disk I/O taking place at all.

Nevertheless I can merge the changes to main if you want, or you might want to fiddle around a bit more. And ofc you can open a new PR anytime.

Hical61:
Could you please help me run again and check the data? Thank you
/benchmark -f hical --save

Hical61:
@Kaliumhexacyanoferrat  Thank you very much for your suggestion. The data is very good and can be merged now

Kaliumhexacyanoferrat (approved):

PR #1059, opened 2026-07-29
Try graalvm image on ring-http-exchange

ruroru:


ruroru:
`/benchmark -f ring-http-exchange

PR #1061, opened 2026-07-30
feat(genhttp-11-kestrel): Add Kestrel engine

Kaliumhexacyanoferrat:


Kaliumhexacyanoferrat:
Interestingly this is faster than ASP.NET Core Minimal API in some of the benchmarks, even though it is built on top of ASP.NET Core and not on raw Kestrel. Great results.

MDA2AV (approved):

Issue #1066, opened 2026-08-04
Where are the JavaScript frameworks in http-arena.com?


nigrosimone:
Hi guys,
Maybe I'm being stupid, but I can't find any JavaScript/Node.js/TypeScript frameworks on https://www.http-arena.com/

Is there anything that needs to be enabled on the filters?

MDA2AV:
Hi, there are typescript (TS) elysia and hono.

Entries are community submitted, unfortunately js seems to not be very incident yet

Kaliumhexacyanoferrat:
To add on this: HTTP Arena relies on contributions by people knowing how to write production grade code for the frameworks to be tested, so framework entries should be done by someone who knows and uses this framework regularly. Therefore we cannot automatically generate them via coding agents or something. So there was just no interest from the JS community to participate (yet).

nigrosimone:
Ok, but I don't find neither "ultimate express" (it's implemented)

MDA2AV:
> Ok, but I don't find neither "ultimate express" (it's implemented)

The implementation was not satisfactory so the results were removed, we can add it once someone reviews and improves it

nigrosimone:
I've added some JS frameworks: Express ([#1068](https://github.com/MDA2AV/HttpArena/pull/1068)), Fastify ([#1069](https://github.com/MDA2AV/HttpArena/pull/1069)), and a fix for the existing Ultimate Express entry ([#1070](https://github.com/MDA2AV/HttpArena/pull/1070)).

Disclaimer, since entries should come from people who actually use these frameworks:

- Fulmine: **I'm the maintainer**
- Ultimate Express: I'm a top contributor
- Express: I'm an expert user (Fulmine and Ultimate Express are Express like framework)
- Fastify: I know it a little, but I'm a top contributor to fast-json-stringify, which Fastify uses for JSON serialization

I'm not affiliate with Ultimate Express, Express and Fastify, neither maintainer has authorized me. I did my best to implement them correctly. As a maintainer of one of the frameworks, I ask reviewers to pay closer attention to PRs for competing frameworks; I want a level playing field where no doubts can be raised regarding the legitimacy of the implementation.

nigrosimone:
Done

Kaliumhexacyanoferrat:
Thank you very much for your contributions!

PR #1068, opened 2026-08-04
Add express framework

nigrosimone:
## Description

Adds **Express 5**, the most widely used JavaScript web framework, which #1066 noted is missing from the arena.

- `meta.json`: JS / flagship / standard, engine `nodehttp`, subscribed to baseline, pipelined, limited-conn, json, json-comp, upload, api-4, api-16, async-db and static
- `Dockerfile`: two-stage build on `node:26-trixie-slim`
- `app.js`: plain Express 5 on `node:http`, cluster module for multi-core scaling, no body parser mounted (the two POST endpoints read the stream themselves)
- Compression is the `compression` middleware and static files are `express.static` serving from disk, both with their default settings, per review

Functional checks were run locally in the container on a mapped port: all endpoints answer with the expected bodies, json-comp is negotiated by the middleware and nothing is compressed without Accept-Encoding. The full `validate.sh` needs host networking, which does not reach a Windows host, so the complete validation rides this PR's CI.

Refs #1066


frameworks/express/app.js
```diff
@@ -0,0 +1,210 @@
...
+            if (count > datasetItems.length) count = datasetItems.length;
+            const m = parseInt(req.query.m) || 1;
+            const items = datasetItems.slice(0, count).map(d => ({
+                id: d.id, name: d.name, category: d.category,
+                price: d.price, quantity: d.quantity, active: d.active,
+                tags: d.tags, rating: d.rating,
+                total: d.price * d.quantity * m
+            }));
+            const body = JSON.stringify({ items, count });
+            // json-comp profile: negotiated per request, nothing without Accept-Encoding
```
Kaliumhexacyanoferrat: For `standard` we must use the default compression middleware instead of a hand-roled version.

frameworks/express/app.js
```diff
@@ -0,0 +1,210 @@
...
+    const MIME_TYPES = {
+        '.css': 'text/css', '.js': 'application/javascript', '.html': 'text/html',
+        '.woff2': 'font/woff2', '.svg': 'image/svg+xml', '.webp': 'image/webp', '.json': 'application/json',
+    };
+
+    // Pre-load static files with pre-compressed variants
+    const staticFiles = {};
+    try {
+        for (const name of fs.readdirSync('/data/static')) {
+            if (name.endsWith('.br') || name.endsWith('.gz')) continue;
```
Kaliumhexacyanoferrat: Custom handling for pre-compressed files is not allowed in `standard`.

frameworks/express/app.js
```diff
@@ -0,0 +1,210 @@
...
+if (cluster.isPrimary) {
+    const numCPUs = getCPUCount();
+    for (let i = 0; i < numCPUs; i++) cluster.fork();
+} else {
+    const express = require('express');
+    const fs = require('fs');
+    const zlib = require('zlib');
+    // level 1: the arena measures throughput of compressed JSON, and the payloads are small
+    // enough that a higher level buys bytes nobody counts
+    const GZIP_OPTS = { level: 1 };
```
Kaliumhexacyanoferrat: Default for production apps would be `6`, for Brotli `11`.

frameworks/express/app.js
Kaliumhexacyanoferrat: Also loading files into memory is only allowed in `tuned`.

Kaliumhexacyanoferrat (changes requested):
There are some things that are allowed in `tuned` but not in `standard`.

nigrosimone:
> There are some things that are allowed in `tuned` but not in `standard`.

done by https://github.com/MDA2AV/HttpArena/pull/1068/commits/bbe098885265c8137ceda247cf96a2684fc023b6  entries now use the stock middleware with default settings

Kaliumhexacyanoferrat:
Looks clean, thanks!

/benchmark -f express --save

MDA2AV (approved):

PR #1069, opened 2026-08-04
Add fastify framework

nigrosimone:
## Description

Adds **Fastify 5**, the second JavaScript framework #1066 found missing.

- `meta.json`: JS / flagship / standard, engine `nodehttp`, subscribed to baseline, pipelined, limited-conn, json, json-comp, upload, api-4, api-16, async-db and static
- `Dockerfile`: two-stage build on `node:26-trixie-slim`
- `app.js`: Fastify 5 with logging off and the cluster module for multi-core scaling; the default body parsers are replaced by a catch-all that hands the raw stream through, since the POST endpoints only count or sum what arrives
- Compression is `@fastify/compress` and static files are `@fastify/static` serving from disk, both registered with their default settings, per review

Functional checks were run locally in the container on a mapped port: all endpoints answer with the expected bodies, json-comp is negotiated by the plugin and nothing is compressed without Accept-Encoding. The full `validate.sh` needs host networking, which does not reach a Windows host, so the complete validation rides this PR's CI.

Refs #1066


Kaliumhexacyanoferrat:
Basically the same issues that we have with the `express` implementation: custom compression, static files in memory, wrong compression levels.

nigrosimone:
> Basically the same issues that we have with the `express` implementation: custom compression, static files in memory, wrong compression levels.

done by https://github.com/MDA2AV/HttpArena/pull/1069/commits/2d7ca26c39ed738d1a821ef69c3a81138eeb7abf  entries now use the stock middleware with default settings

MDA2AV (approved):

PR #1065, opened 2026-08-03
Add fulmine framework

nigrosimone:
## Description

Adds **fulmine** (`fulmine.js`), an Express 5 compatible framework built on µWebSockets.js

- `meta.json`: JS / emerging / tuned, engine `uwebsockets`, subscribed to baseline, pipelined, limited-conn, json, json-comp, upload, api-4, api-16, async-db and static
- `Dockerfile`: two-stage build on `node:26-trixie-slim`
- `app.js`: the arena endpoints on the Express 5 API, cluster module for multi-core scaling
- `fulmine.js` is pinned `^5.0.0`, so rebuilds pick up the latest 5.x

The entry was validated in container with `./scripts/validate.sh fulmine`

Ref https://github.com/MDA2AV/HttpArena/issues/1066

PR #1070, opened 2026-08-04
Fix ultimate-express entry: json-comp negotiation was missing

nigrosimone:
## Description

#1066 says the ultimate-express results were removed because the implementation was not satisfactory. Reviewing the entry, the concrete defect is that `meta.json` subscribes to **json-comp** but `/json/:count` never negotiated `Accept-Encoding`: every response went out as plain JSON, which the json-comp validation refuses.

- `app.js`: `/json/:count` now negotiates gzip (level 1) and brotli (quality 3) per request, and still sends plain JSON without `Accept-Encoding`, matching the other entries
- `meta.json`: the description said *single-process threading* but the code scales with the cluster module; reworded
- `package.json`: renamed from `httparena-express` to `httparena-ultimate-express`

Functional checks were run locally in the container on a mapped port: json-comp now negotiates both encodings and the gzip body round-trips. The full `validate.sh` needs host networking, which does not reach a Windows host, so the complete validation rides this PR's CI.

Refs #1066


Kaliumhexacyanoferrat:
This is marked `tuned`, so the points from the other PRs are fine here - but you might want to align the implementation here as well while you are already at it?

nigrosimone:
Dockerfile, dependencies and README are now aligned with the other node entries (592fd01).

On the implementation itself, just to be sure I read you right: align meaning switching to the official `compression` middleware here as well? I kept this one hand-tuned on purpose, since the entry is `tuned` and I'd rather have ultimate-express at its best: I maintain a competing framework, so slowing this entry down is the last look I want. But if you prefer the middleware version for consistency or maintainability, happy to switch it.

Kaliumhexacyanoferrat:
Sorry, my fault - reading too hastily.

/benchmark -f ultimate-express --save

MDA2AV:
Hi, all looks good except the static test, you are caching the files data in memory, we are changing rules so that tuned entries can no longer do that, can you update this and fulmine entries?

nigrosimone:
Done for both entries: ultimate-express here (ecf22a3) and fulmine on #1065 (66610ff). No file data lives in memory any more: startup scans only the names, the existing pre-compressed variants and the content types, and the bytes are read from disk on every request. The negotiation itself (brotli or gzip variant by Accept-Encoding) is unchanged.

nigrosimone:
Merged `main` into the branch, re-requesting:

/benchmark -f ultimate-express --save

MDA2AV (approved):

PR #1073, opened 2026-08-04
Bump fulmine to 5.1.3

nigrosimone:
## Description

fulmine 5.1.3 is a performance-only line of releases since the 5.1.1 the saved results measured: cork elision inside uWS's own corked window, deferred stream end on bodyless requests, fewer per-request allocations, and a stat-validated file cache behind sendFile. The pin moves to `^5.1.3` so the entry declares its real minimum.

This is the last one, guys, I won't bother you anymore :D

Requesting a refresh of the saved results when the runner has room:

/benchmark -f fulmine --save


nigrosimone:
benchmark -f fulmine --save

MDA2AV (approved):

MDA2AV:
No problem, run as many as needed.

PR #1074, opened 2026-08-05
Tune the fulmine async-db path

nigrosimone:
## Description

Four changes to the async-db path, each with its reasoning:

- **Pool sized under `max_connections`**: the flat `max: 4` per worker, times ~64 workers, lands exactly on Postgres's 256 and every request pays the contention. Now `floor(240 / cores)` per worker, the same arithmetic the elysia entry documents for the same reason.
- **Named prepared statement**: `pg`'s parameterized query alone re-parses on every execution; the `name` field makes it Parse once per connection, Bind/Execute afterwards.
- **`pg-native` preferred, JS client as fallback**: the libpq bindings measure fastest of the node drivers at this profile's row counts (1-50 rows) in [nigrosimone/postgres-benchmarks](https://github.com/nigrosimone/postgres-benchmarks), the benchmark node-postgres itself references: median 206us vs 233us (`pg`) and 225us (`postgres`) at 1 row, still ahead at 100, converging at 500. The Dockerfile builds it against libpq-dev and ships libpq5.
- **Dead code removed**: the `/db` SQLite route is requested by no profile, so `better-sqlite3` was a native module loaded in every worker for nothing.

Validated in container locally (all endpoints, json-comp negotiation, static from disk). Asking for a test run first; if async-db and api-4/16 move as expected I'll ask for a `--save`.


nigrosimone:
benchmark -f fulmine

nigrosimone:
async-db +9.7%, api-16 +4.2%, static +2.5-3.2%, memory down where the native module left, everything else within noise: exactly the expected shape. Requesting the save:

/benchmark -f fulmine --save

Kaliumhexacyanoferrat:
Just a short hint; You can also run with `--save` on first attempt - if the results are fine we can just merge - if not, we can re-run and re-save with another commit.

nigrosimone:
Good to know, thanks: will go straight to --save next time.

nigrosimone:
fulmine 5.1.4 is out (query parser and abort-registration work, ~12% less CPU per request on the mixed-API shape), the pin now takes it. Straight to save as suggested:

/benchmark -f fulmine --save

MDA2AV (approved):

PR #1076, opened 2026-08-05
Bump fulmine to 5.1.7

nigrosimone:
Pin bump for the new fulmine release. Since the 5.1.4 run, the framework collects request bodies natively with the size limit enforced before any byte reaches JavaScript, and routes whose handlers provably never read a request header skip the header copy entirely (a registration-time source analysis grants it, and anything it cannot prove keeps the copy). The entry itself is unchanged.


nigrosimone:
C:/Program Files/Git/benchmark -f fulmine --save

nigrosimone:
benchmark -f fulmine --save

nigrosimone:
Added the three WebSocket echo profiles to this PR, since it is the same entry under review: `/ws` echoes each frame back through µWebSockets.js's own WebSocket server (`app.uwsApp.ws`), which fulmine exposes directly. Every connection performs the real µWS upgrade handshake, no caching of the 101 across connections. Smoke-tested the container locally: upgrade plus repeated text echoes, and the HTTP routes are untouched.

/benchmark -f fulmine --save

nigrosimone:
Two more profiles for the same entry, both smoke-tested locally against real sidecars (Postgres seeded with the schema, Redis, self-signed certs):

- **json-tls**: the same `/json/:count` handler served over µWS's native TLS on port 8081 (`SSLApp` under fulmine's `uwsOptions`), certs read from `/certs`. Verified the response shape and the `?m=` multiplier over HTTPS.
- **crud**: the four endpoints per the spec, named prepared statements on the pool already tuned for async-db, and the Redis sidecar as the cache backend since the entry runs one process per core (cached value is the pre-serialized body, `PX 200`). Verified list shape, `X-Cache: MISS` then `HIT` on the same connection, upsert create with 201, update with cache invalidation (next read is a MISS with the new data), and 404 on a missing id.

/benchmark -f fulmine --save

MDA2AV (approved):

PR #1085, opened 2026-08-07
Bump fulmine to 5.5.0

nigrosimone:
Pin bump for the fulmine entry, from `^5.2.0` to `^5.3.0` ([changelog](https://github.com/nigrosimone/fulmine.js/blob/main/CHANGELOG.md)).

The release is mostly compatibility work: 27 fixes, nearly all of them found by a differential fuzzer that builds random applications and compares the answers against real Express. Nothing there should move a benchmark row.

Four changes do touch the request path, each measured on our own rig against the previous revision, so I have no prediction for yours:

- the response arms its two internal listeners by writing the event map directly instead of calling `on()` twice, -6.9% CPU per request on a hello-world
- the request no longer builds a `Readable` until something asks for one, -5.2% and -5.5% over two runs
- an IPv4-mapped peer address is read straight out of the bytes rather than through the general IPv6 formatter, which shows on anything that reads `req.ip`
- a chain steps over a body parser when the request declared no body and the verb reads none, which removes a hop rather than shortening one

No entry files change other than the pin.


nigrosimone:
C:/Program Files/Git/benchmark -f fulmine --save

nigrosimone:
Sorry, my shell rewrote the leading slash in the comment above into a Windows path. The command I meant:

/benchmark -f fulmine --save

nigrosimone:
Pushed a second commit, so this PR is now the pin bump plus a change to the crud cache. Happy to split it if you would rather review them apart.

**`ioredis` → `@redis/client`, with the cache it can keep in the process.** The client is created with `RESP: 3` and `clientSideCache: { ttl: 200, maxEntries: 4096, evictPolicy: 'LRU' }`. A cache HIT is then answered inside the worker with no round trip, and Redis pushes the invalidation when a key stops being true, so the sidecar still owns correctness. The cache is bounded and LRU on purpose: there is one process per core, and 64 unbounded caches would be the working-set problem the sidecar was introduced to solve.

Two details that needed care:

- `connect()` is a promise here, where ioredis connected in the background. Until it settles the in-process map answers, which is what already happened when the harness provided no sidecar: a cache that is not ready is a miss, never an error. Verified with a client pointed at a dead port: the fallback holds and nothing is left unhandled.
- a write drops the key from this process too, and after the write rather than before it. The invalidation Redis pushes is a round trip away, and a read on the same connection immediately after the write must not beat it home.

`enableAutoPipelining` has no counterpart because node-redis has no such switch: it queues commands and flushes them on `setImmediate`, so pipelining is not optional there.

**The crud read path built its headers per request.** `res.set({ ...SERVER_HDR, 'x-cache': 'HIT' })` is two allocations on the busiest route the entry has. They are two constants now.

Smoke tested in Docker against postgres 17 and redis 7 with the items table seeded: read MISS then HIT on the same connection, 404 for an id that is not there, list by category, POST upsert 201, and PUT followed by a read that comes back MISS with the new values, which is the invalidation doing its job. What I could not run here is `scripts/validate.sh`: it needs `--network host`, which Docker Desktop does not expose to a Windows host.

nigrosimone:
Thanks for running it. One row in that result is mine and it is a regression, so I have taken the change back.

`crud` at 4096 connections, against the previous run of this entry:

| | before | after |
| --- | ---: | ---: |
| rps | 352,481 | 210,973 |
| avg latency | 11.56ms | 19.41ms |
| p99 | 21.20ms | 43.90ms |
| cpu | 4502% | 3236% |
| memory | 8.1GiB | 11.2GiB |
| non-2xx | 0 | 0 |

The CPU is the interesting column: it went **down** while latency doubled, and nothing failed, so the
processes were waiting rather than working. That is a bottleneck that moved out of this entry and into
Redis. Server-assisted client-side caching makes Redis keep an invalidation table per key per client
and push a message to all 64 of them whenever one writes, and this test writes on every cache miss.

The last commit puts `ioredis` back exactly as it was, with the reason written next to it so nobody
tries it again. Two things from those commits are kept because they are independent of the client:
the two crud response-header objects are built once instead of being spread per request, and the body
reader joins its chunks as buffers so a UTF-8 character split across two chunks survives.

Sorry for the wasted run. Whenever you can re-run it, `crud` should be back around 352k; the rest of
the table was already fine, with `pipelined`, `json`, `json-tls`, `api-16` and `async-db` up 3-5%
from the 5.3.0 bump itself.

nigrosimone:
Bumped to 5.4.0, which is on npm now, so this PR is one thing to approve rather than two: the
`ioredis` revert and the new version get measured together.

What is in it that this benchmark can see:

- `crud` should return to around 352k. That row is the reason for the revert, and the previous run's
  40% drop is explained in the comment above.
- Four compatibility fixes, none of which change the shapes this benchmark exercises: a mount no
  longer answers past a layer written before it, `res.format`'s 406 leaves the route as Express's
  does, `req.query` hands back a fresh object per read as Express's getter does, and
  `res.getHeader("content-length")` answers instead of coming back empty.
- Two small ones on the response path: it no longer builds its Writable state until something asks,
  measured at 45ns a response, and the peer-address read no longer reopens its window every 100000
  requests.

Nothing here should move the other rows outside noise. If any of them does, I would rather hear it
than not, since this rig is steadier than mine.

MDA2AV:
there is no "lightning" framework, I think it somehow translated from italian to english lol

MDA2AV (approved):

nigrosimone:
Fulmine = Lightning
:D

nigrosimone:
Bumped to 5.4.1. The last run's -4..-9% on the query-carrying rows was real, it was mine, and this release removes it.

Expected on re-run: baseline, limited-conn, json, json-tls, json-comp and api-4 back to their 5.3.0 numbers, crud staying at ~357k, everything else unchanged.

nigrosimone:
Merged `upstream/main` in, so the branch is mergeable again and `/benchmark --save` should start. The three commits it picked up (static-tls, the pipelined reference-only change, the new PR commands) touch nothing under `frameworks/fulmine`, and the entry is unchanged at `^5.4.1`.

Same expectation as before: `crud` around 357k, and the query-carrying rows (baseline, limited-conn, json, json-tls, json-comp, api-4) back to their 5.3.0 numbers now that the copy-per-read is gone.

nigrosimone:
Bumped again, to 5.5.0, which is on npm. Nothing in it changes what this benchmark exercises: it adds `res.flushHeaders()` and fixes two TypeScript declarations, all three found by running an Angular SSR application on it. The performance-relevant change is still the one from 5.4.1, the `req.query` fix.

Expectation for the run is unchanged: `crud` around 357k, and baseline, limited-conn, json, json-tls, json-comp and api-4 back to their 5.3.0 numbers.

The branch is mergeable with main as of the merge two commits back.

nigrosimone:
It's good for me! 🎉

PR #812, opened 2026-06-06
[Rust] Add WTX - gRPC

c410-f3r:
## Description

`WTX` also has a `PostgreSQL` client but related benchmarks only exist for HTTP/1.1. Therefore, this is the last protocol.

PR #1095, opened 2026-08-09
Add Cardigan Java io_uring engine

phjeltman:
## Description

First, thank you for picking up the benchmarking torch after TechEmpower.

This PR adds Cardigan, a Java HTTP engine built on io_uring, Panama FFI and virtual threads. The submission uses the published `dev.cardigan:cardigan:0.1.0-alpha1` artifact fresh off Maven Central. It covers HTTP/1.1, HTTP/2 over TLS, h2c, JSON over TLS, and unary and server-streaming gRPC. The entries share one implementation and subscribe to non-overlapping profiles.

One caveat: `validate.sh` currently has no gRPC-specific readiness or endpoint validation, so it fails before exercising gRPC-only entries. I verified Cardigan's unary and server-streaming variants directly with `ghz`, over both cleartext and TLS. [WTX encountered the same validator limitation in #812](https://github.com/MDA2AV/HttpArena/pull/812).

---

| Command | Description |
|---------|-------------|

```
/benchmark -f genhttp-11 --compare genhttp
```

---

<details>

```bash
```

</details>


MDA2AV:
gRPC tooling is currently broken on validation, and tbh ghz as a load generator is also quite odd, grpc stream tests are still wip

PR #1097, opened 2026-08-10
Add Proxygen and Proxygen Coro engines

lnicco:
Early review requested. This draft has been rewritten as an engine-only two-commit stack on the current main branch.

## Summary

- add regular Proxygen with HTTP/1.1, HTTP/1.1 TLS, h2c, HTTP/2 TLS, HTTP/3 QUIC, and RFC 6455 WebSockets
- add proxygen-coro as a self-contained coroutine engine with the same 18 test profiles
- group both variants under the proxygen engine key
- track ghcr.io/facebook/proxygen/base:latest for both builders
- keep the shared HttpArena harness unchanged

The regular entry uses one process with classic HTTPServer for every TCP listener and HQServer only for HTTP/3. The coroutine entry also uses one process: its four TCP listeners share one worker pool, while HTTP/3 uses the second pool required by the current coroutine API. Both use 1 MiB / 10 MiB HTTP/2 flow control, 1024 concurrent streams, and upstream benchmark-oriented QUIC batching and write-path settings.

## Validation

- no-cache builds of both self-contained images against the current base:latest image
- HTTP/1.1, HTTP/1.1 TLS, h2c, and HTTP/2 endpoint probes passed for both engines
- direct HTTP/3 baseline probes passed for both engines
- extended WebSocket validation: 20/20 for regular and 20/20 for coro
- four-CPU topology check: one process and two four-thread transport pools per engine
- graceful SIGTERM shutdown: 0.12s regular, 0.15s coro
- metadata, JSON, shell syntax, ShellCheck, and patch checks passed

### DIFF
--- a/frameworks/proxygen-coro/ArenaCoroServer.cpp
+++ b/frameworks/proxygen-coro/ArenaCoroServer.cpp
@@ -1,0 +1,974 @@
+#include <algorithm>
+#include <array>
+#include <cctype>
+#include <charconv>
+#include <chrono>
+#include <csignal>
+#include <cstdint>
+#include <deque>
+#include <fstream>
+#include <future>
+#include <iostream>
+#include <iterator>
+#include <limits>
+#include <list>
+#include <memory>
+#include <set>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <thread>
+#include <utility>
+#include <vector>
+
+#include <folly/ScopeGuard.h>
+#include <folly/base64.h>
+#include <folly/dynamic.h>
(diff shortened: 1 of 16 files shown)


MDA2AV:
Hi, let me know if you need more runs.

lnicco:
thanks @MDA2AV ! I am handing this over to a teammate and we may need a few more iterations. But we'll publish only when everything looks good. Sorry for the noise

lnicco:
replaced by https://github.com/MDA2AV/HttpArena/pull/1435

PR #1098, opened 2026-08-11
Bump fulmine to 5.5.2

nigrosimone:
Pin bump for the fulmine entry, from `^5.5.0` to `^5.5.2` ([changelog](https://github.com/nigrosimone/fulmine.js/blob/main/CHANGELOG.md)), plus three changes to the entry.

In the release, measured here against the previous revision:

- a `use` with no path ran a regex on every hop for a mount prefix that was always 0. Removing it is -16.8% CPU per request on a chain of 100 middlewares, 23.1k to 27.6k req/s
- the ETag hashed a copy of the body instead of the body: -7.7% on a 35 byte body, -41% on 54 KB
- the generic handler drops a promise pair nobody awaited, and a response no longer allocates its chunk queue or its flush closure unless it streams

In the entry:

- `/static` is served on the TLS listener as well, so it subscribes to `static-tls`, same handler as the plaintext one.
- a stock Caddy edge in `proxy/`, so it subscribes to `gateway-64` and `gateway-h3`. Caddy terminates TLS and serves `/static/*` from disk, the server keeps the three dynamic endpoints it already had, reached over loopback h1. I ran both stacks locally against the repo certs: h2 and h3 both negotiate, `/static` comes from the proxy with the right content types and a 404 for a missing file, and `/baseline2`, `/json/N` and `/async-db` all answer through it.
- `json-comp` prefers brotli, which it never reached before because `Accept-Encoding: gzip, br` matched gzip first. On these payloads q3 is 12% smaller than gzip level 1 for about 24us more per request, and with the score squaring the body size that should come out ahead, but it is a model and not your rig, so the run decides.

/benchmark -f fulmine --save


nigrosimone:
benchmark -f fulmine --save

nigrosimone:
The `json-comp` row reads -16% on RPS, which is the brotli change and is expected rather than a regression: that profile scores `rps * (minBpr/myBpr)^2`, so the bytes count twice over.

Bytes per response went 1562 to 1374, so the size term is 1.29 against 0.84 on rps. Against the field's current minBpr of 1107 the profile score goes 16.6 to 18.6.

Still well short of the 1107 the leader ships, so there is more left there, but brotli above q3 gets very slow on payloads this small and stopped paying for itself.

MDA2AV (approved):

PR #1100, opened 2026-08-11
fulmine: add the production-stack entry

nigrosimone:
Subscribes the fulmine entry to `production-stack`. Four services as the rules ask: a stock Caddy edge, Redis, the shared authsvc built as-is, and the server.

The edge terminates TLS and h2, serves `/static/*` off disk, sends `/public/*` straight through, and puts `/api/*` behind `forward_auth` against the authsvc, copying `X-User-Id` on. Nothing in the server looks at a token.

On the server side most of it was already there for `crud`: the cache-aside read is now registered under both `/crud/items/:id` and `/api/items/:id`, and the JSON route under `/public/json/:count`. New are `/public/baseline`, the `POST /api/items/{id}` that answers 204 and invalidates, and `/api/me`, which reads the user by `X-User-Id` with a 30 second TTL.

I ran the stack locally against the repo certs and seed, and walked the validation checks by hand: h2 negotiates, `/static` comes from the edge, `/public/baseline` gives 55, `/public/json/25` has 25 items with the full schema and correct totals, `/api/*` is 401 without a token and with a bogus one on both GET and POST, `/api/items/1` returns id 1, cache-aside reads MISS then HIT, POST answers 204 and the next read is MISS again, and `/api/me` returns user 42.

/benchmark -f fulmine -t production-stack --save


nigrosimone:
The `production-stack` rows landed at 19.5k, which is well off the 211k the leading entry does, and the CPU column says why: 1537% and 1577% against the 1600% my edge's sixteen logical CPUs allow. The edge was pegged and the other three services sat idle.

I had the split backwards. The nginx entry gives its proxy 30 logical CPUs and its server 24; I had given the edge 16 and the server 38, on a profile where the edge terminates TLS and h2 for every request and serves `/static` off disk on top. The edge now takes 32 and the server 22.

I also suspected the `forward_auth` hop was dialling the verifier per request and added a keepalive transport for it, then measured before and after: 200 `/api` requests leave two established connections either way, so Caddy was already pooling and the change bought nothing. Dropped it rather than carry a fix for a problem that wasn't there.

Happy to sweep `EDGE_CPUSET`/`SERVER_CPUSET` further if the next run still leaves headroom.

/benchmark -f fulmine -t production-stack --save

nigrosimone:
It's good for me

MDA2AV (approved):

PR #1101, opened 2026-08-11
fulmine: subscribe to fortunes

nigrosimone:
Subscribes the fulmine entry to `fortunes`. It is reference-only, so it moves no score; what it adds is the one column the row was missing, and the only profile here that puts a template engine under load.

Rendered with EJS through `res.render` and a template file rather than the string building the tuned rules would also allow, since hand-rolling it would measure something the profile is not asking about. The handler queries the 200 rows, appends the runtime one in memory, sorts ordinally and renders per request.

Checked locally against the repo seed: `text/html; charset=utf-8`, doctype present, 202 `<tr>` (one header plus the 201 rows), the runtime row present, `&lt;script&gt;` escaped with no raw `<script>alert` anywhere, and a 24,536 byte body. The sort is right too: the escaped `<script>` row comes first and the Japanese one last, which is what ordinal order gives.

/benchmark -f fulmine -t fortunes --save


nigrosimone:
Ok!

MDA2AV (approved):

PR #1102, opened 2026-08-12
fulmine: move the entry to standard mode

nigrosimone:
The entry was tuned only because the framework was missing two things. Both shipped in fulmine.js
5.6.0 today, so it can run under the standard rules instead.

- `express.compression()` replaces the hand-rolled `zlib.brotliCompressSync`/`gzipSync` on the json
  route. It is the framework's own middleware and takes the compression module's options.
- `express.static(dir, { preCompressed: true })` replaces the `fs.readFile` with the MIME table and
  the `.br`/`.gz` suffix lookup. Serving the twins is a documented option, not custom logic.
- `express.json()` replaces the hand-rolled body reader, with an error handler for the empty 400.
- `app.set("file cache", false)` turns off the small-file cache, so every request reads from disk.
- `app.set("etag", false)` stays. It is recommended for an API whose responses are never revalidated
  in the framework's own production guide, under Performance tips.

Checked locally against the validation rules: the files come back at their exact size without
`Accept-Encoding`, decompress to that size with `br;q=1, gzip;q=0.8`, the woff2 and webp are left
alone, a missing file is a 404 and the content types are unchanged.

The numbers will move: `express.static` also sends ETag, Last-Modified, Accept-Ranges and
Cache-Control, about 125 bytes more per response, and the static profile weighs bytes.


nigrosimone:
👍

MDA2AV (approved):

PR #1104, opened 2026-08-12
Bump fulmine to 5.7.0

nigrosimone:
Two changes since the run that came with #1102, both of them where that run lost ground:

- The compression middleware no longer wraps the response when the request cannot take a compressed
  body. The json and json-tls profiles send no `Accept-Encoding`, so they were paying for a
  compressor that was never going to run.
- `preCompressed` costs one `stat` per request instead of two or three: the twin is looked for
  before the file rather than after it, a type that is already compressed is not looked up at all,
  and which twins a path has is remembered for a second.

Measured here against the hand-rolled route the entry used to carry, on a rotation over a mixed
directory, that is -35% before and parity after. This machine is Windows on 8 cores, where a `stat`
costs far more than it does on your rig, so the runs that matter are yours: `static`, `static-tls`,
`json` and `json-tls` are the ones worth a look.


nigrosimone:
👍

MDA2AV (approved):

PR #1106, opened 2026-08-13
Bump fulmine to 5.9.0

nigrosimone:
5.9.0 adds a `cluster` option to the framework, so the entry no longer carries the fork itself:
`express({ cluster: 'auto' })` forks one worker per usable core and each worker binds the port with
uWS's shared flag, which is what the entry was doing by hand. The worker count reads the cgroup
quota before the machine, so it is the container's cores and not the runner's.

Same processes, same shared port: nothing about how it serves should move. The entry lost its
cluster boilerplate, which is the point, since the file is read as documentation of the framework.

Everything else is unchanged. The diff looks large only because the body is no longer indented
inside an `else`.


nigrosimone:
👍

MDA2AV (approved):

PR #1117, opened 2026-08-14
fulmine: 5.11.0, and /pipeline written so it compiles

nigrosimone:
fulmine.js 5.11.0, and one change to the entry.

`/pipeline` was going through the ordinary router rather than through the declarative response this entry's README claims for it. The compiler reads the handler's source and cannot see into a closure, so `res.set(SERVER_HDR)` kept it off that path. Written out as a literal it is compiled at listen() and answered by uWS without entering JavaScript.

Standard mode as before: the same documented APIs, no new setting, and nothing the entry precomputes itself. If you read a compiled handler as a pre-computed response body under the workload rules, say so and I will put it back.

5.11.0 also gives a compiled route a Content-Length, so its framing now matches Express, and it is what the entry runs with.


nigrosimone:
benchmark -f fulmine --save

MDA2AV (approved):

PR #1121, opened 2026-08-14
Re-validate all frameworks

Kaliumhexacyanoferrat:
Whitespace change for every readme file to re-trigger the validation on all frameworks.


Kaliumhexacyanoferrat:
Validation still fails if one matrix entry fails - fixing that in #1122

ruroru:
Good idea, but its missing frameworks without README.md.  Maybe you can try creating empty README.md in that case

Kaliumhexacyanoferrat:
## Validation re-run triage

44 of the 159 frameworks failed `validate` in this run. I went through every failing job's logs and grouped them by root cause below. Nothing here is guesswork — each entry is backed by the actual container/build logs from [this run](https://github.com/MDA2AV/HttpArena/actions/runs/31810448715).

### 🔧 Validation-script / CI-infra issues (not framework bugs)

These three cover **17 of the 44 failures** and are false positives caused by gaps in `scripts/validate.sh` or by the recent move to GitHub-hosted runners — the frameworks themselves are very likely fine.

**1. gRPC test profiles are unimplemented in `validate.sh`** — 8 frameworks: `wtx-grpc`, `wtx-grpc-tls`, `tonic-grpc`, `grpc-go`, `spring-jvm-grpc`, `quarkus-jvm-grpc`, `cardigan-grpc`, `cardigan-grpc-tls`.
`scripts/validate.sh` has no reference to `grpc`/`unary-grpc`/`stream-grpc` anywhere: no readiness-probe skip (there's one for ws-only frameworks, none for grpc-only), and no actual grpc assertions. Effect: the script waits for a plain-HTTP `/baseline11` response that a grpc-only server never serves, so it always times out with `FAIL: Server did not start within 30s` — even though the build succeeded and the server is presumably up. The `-tls` variants have it worse: `needs_h2`/`needs_h1tls` never check for `unary-grpc-tls`/`stream-grpc-tls`, so `$CERTS_DIR` never gets mounted into the container. `wtx-grpc-tls` and `cardigan-grpc-tls` both crash on startup trying to open a cert that was never mounted (`IoError(Os { code: 2, kind: NotFound })` / `loading certificate chain: ... No such file or directory`).

**2. Gateway/production-stack compose profiles hardcode CPU ranges for a many-core benchmark host** — 7 frameworks, **100% failure rate** for anyone subscribed to these profiles: `trillium`, `trillium-tuned`, `aspnet-minimal_caddy`, `aspnet-minimal_nginx`, `fulmine`, `sark-gateway`, `sark-production`.
The `compose.gateway*.yml` / `compose.production-stack.yml` files pin containers to specific cores (e.g. `cpuset: "16-31,80-95"`), assuming a ≥96-core host. GitHub Actions runners only expose 4 cores: `Error response from daemon: Requested CPUs are not available - requested 16-31,80-95, available: 0-3`. Every single framework that subscribes to `gateway-64`/`gateway-h3`/`production-stack` failed this way (`trillium-tuned` hit the build timeout below before it got this far, but subscribes to the same profiles and would hit this too). This isn't a per-framework problem — it needs a CI-appropriate cpuset (or making it configurable) in the compose files/validate.sh itself.

**3. 300s `VALIDATE_TIMEOUT` is too short for heavy from-source builds on standard runners** — `trillium-tuned`, `humming-bird`, `aspnet-minimal-iouring`.
`trillium-tuned` and `humming-bird` were both still mid-compile (cargo / swiftc) when the watchdog killed them at 300s. `aspnet-minimal-iouring` builds a patched CoreCLR from source — it was at file 230/2697 when killed. Combined with `docker build --no-cache` (validate.sh always disables layer caching) and no persistent build cache across GH Actions runs, these can't realistically finish in the window. Needs either a longer timeout for known-heavy frameworks, actual build caching between runs, or bigger runners — not a framework code issue.

*(Aside, unrelated to `validate.yml`: the `notify` check on this PR fails for an unrelated reason — `notify-maintainers.yml` uses `pull_request_target` with a full checkout of the PR head, which GitHub now refuses for cross-repo/fork PRs as a "pwn request" guard. Since this PR is from a fork, maintainer auto-pings for framework changes aren't currently firing. Separate fix, just flagging it since it's in the same neighborhood.)*

### 🐛 Genuine framework issues (maintainer action needed)

- **@enghitalo** — `vanilla-h2c`, `veb`, `vanilla-io_uring`, `vanilla-ws`, `vanilla-epoll`, `pico`, `fasthttp` (7 frameworks, all identical): build fails compiling the pinned V compiler commit from source — `vlib/os/filelock/lib_nix.c.v:7:6: error: C function C.open was already declared with a different signature`. The pinned `vlang/v` commit no longer builds against whatever glibc/headers the base image now resolves to. Needs a compiler commit bump (or pinning the base image digest so it stops drifting).
- **@prothegee** — `zix`, `zix-ws`, `zix-grpc` (3): Dockerfile fetches `https://github.com/prothegee/zix/archive/refs/heads/${ZIX_VERSION}.tar.gz` → `404`, and both the codeberg and git-clone fallbacks fail too. The pinned `ZIX_VERSION` ref doesn't exist upstream anymore.
- **@gdiazlo** — `araara`, `araara-standard` (2): `opam source --dir hcs "hcs.0.16.1"` → `No package matching hcs.0.16.1 found`. Pinned `HCS_VERSION` no longer resolves in the araara opam repo.
- **@joanhey** — `php`: `apt-get install ... php8.5-fpm php8.5-cli php8.5-pgsql` from `ppa:ondrej/php` exits 1. The real apt error is hidden behind `>/dev/null` in the Dockerfile — worth dropping that redirect to see what's actually failing (possibly php8.5 packages not yet available for this PPA/Ubuntu combo).
- **@joanhey** — `workerman-websocket`: `composer install` fails — "the zip extension and unzip/7z commands are both missing" in the build image, so composer can't unpack zip-distributed dependencies.
- **@YanGusik** — `symfony-spawn-franken`: build fails because the `php` binary itself can't load — `libwebpdemux.so.2: cannot open shared object file`. Missing runtime lib in the build image.
- *(no maintainer listed)* — `frankenphp-trueasync`: same missing-lib symptom, but at container startup instead of build time — `frankenphp: error while loading shared libraries: libwebpdemux.so.2`. Possibly shares a base-image lineage with `symfony-spawn-franken` above — worth checking if they're related.
- *(no maintainer listed)* — `h2o-mruby`: server fails to start — `h2o: error while loading shared libraries: libbrotlidec.so.1: cannot open shared object file`. Runtime lib missing from the final image stage.
- **@LazyPanda07** — `web-framework-python`: build "succeeds" (222s) despite a `CMake Error` at `3rdparty/DotNetRuntime/CMakeLists.txt:33` and a failed ninja/cmake_install step partway through. At runtime the resulting `libWebFramework.so` fails to load: `Can't load ... or its dependencies`. The vendored `DotNetRuntime` component isn't building correctly and the final `.so` is broken as a result.
- **@inq** — `sark`: HTTP/2+TLS is entirely non-functional (port 8443 not responding, `HTTP/0.0` negotiated, empty bodies on `/baseline2`) — separate from a real caching bug where `crud cache-aside` returns `MISS` twice instead of `MISS` then `HIT`.
- **@inq** — `sark-h3`: build succeeds, container stays up, but never answers the h2-over-TLS fallback probe on :8443 within 30s (zero log output, so hard to tell if it's a slow QUIC-listener init or a real hang — may be worth checking if it needs more than 30s on constrained CI hardware, or if this is a genuine startup issue).
- *(no maintainer listed)* — `ringzero`: 5 checks fail, all on TCP-fragmented HTTP/1.1 requests (split request line, split before headers, split headers/body, lower-cased headers) — either an empty response or a stale/leaked body from the previous request, indicating the parser doesn't reassemble partial `recv()`s correctly and/or connection state bleeds across requests.
- `genhttp` (my own entry, no need to ping): 47/49 checks pass. The 2 failures are the same TCP-fragmentation pattern as `ringzero` — `split request line` / `split before headers` both return an empty body with `BrokenPipeError`, i.e. the connection is closed before the rest of the fragmented request arrives. I'll fix this myself.
- *(no maintainer listed)* — `node-websocket`: 3 WS checks fail. `text echo`/`multi-message` show matching content in the log but still fail — `scripts/validate-ws.py` requires both the payload **and** the frame opcode to match (`opcode == OP_TEXT and echoed == msg`); the pass/fail message only prints the payload, so a same-looking pass/fail is actually an opcode mismatch (server is very likely echoing text frames back with the wrong opcode). The subsequent `server alive` health check then also fails, so the connection doesn't survive the run either.
- *(no maintainer listed)* — `deno-websocket`: `GET /ws` without an `Upgrade` header returns HTTP 500 instead of a graceful rejection — an uncaught `TypeError: Invalid Header: 'upgrade' header must contain 'websocket'` in `server.ts`.
- **@MDA2AV** — `fleck`: same non-upgrade-GET case returns HTTP 0 — connection reset, no response at all.
- **@MDA2AV** — `dogrider`: `binary echo` fails with the same opcode-vs-payload issue as `node-websocket` above — `validate-ws.py` requires `opcode == OP_BINARY and payload == data`; the byte counts match (256/256) but the message format doesn't distinguish opcode from content mismatches, so this reads as "sent 256, got 256" even though it's actually failing. Very likely echoing binary frames back with the text opcode.
- *(no maintainer listed)* — `flask`: 3 Content-Type checks fail — plain-text responses come back as `text/html; charset=utf-8` instead of `text/plain`. That's Flask's default mimetype for string returns; needs an explicit `mimetype="text/plain"`.
- *(repo-internal, no maintainer listed)* — `aspnet-minimal_caddy`, `aspnet-minimal_nginx`: both Dockerfiles `COPY frameworks/aspnet-minimal/Models.cs` and `AppData.cs`, but those files no longer exist there — they were folded into `Services/`/`Types/` by #1002's refactor of `aspnet-minimal`. Cross-framework `COPY` dependency broke when the source framework was refactored. (These two would also hit the CPU-pinning issue above once the build itself is fixed.)

---

**tl;dr**: 17 of the 44 failures (gRPC profiles, CPU-pinned gateway/production-stack compose, and the 300s build timeout) look like validation-script/CI-infra gaps rather than framework regressions — I'd suggest fixing/relaxing those in `validate.sh` before drawing conclusions about the frameworks they affect. The rest are pinned dependencies that broke upstream (vlang, zix, araara/opam), missing runtime shared libs, and a handful of genuine parser/logic bugs, tagged above by framework with the maintainer to ping where one is listed.

Kaliumhexacyanoferrat:
Rebased onto main with new gRPC validation. Increased timeout to 10 minutes.

prothegee:
Thanks for the info.

Working on the latest version on 0.5.0, will inform and re-bench later.

Kaliumhexacyanoferrat:
@MDA2AV fixed the infra-related issues with the validation, some of them due to my change to GitHub actions - sorry!

- Rebased onto your gRPC fix
- Gateway validation does now null the cpuset instructions by the frameworks
- Set job timeout to 10 minutes

So I would merge this to have this fixes on main.

Kaliumhexacyanoferrat:
@MDA2AV ah, I see this was not merged so the gateway validation still failed on PRs. The approach here was to overwrite the cpuset instructions via docker compose overlays. As you found another way I will then close this here.

MDA2AV:
> @MDA2AV ah, I see this was not merged so the gateway validation still failed on PRs. The approach here was to overwrite the cpuset instructions via docker compose overlays. As you found another way I will then close this here.

Ah.. I was actually wondering what happened to the fix, I just applied the quick workaround to validate gateway directly on the server machine, we can consider this in the future if more frameworks adopt gateway tests

PR #1125, opened 2026-08-14
fulmine: the last production-stack run had authsvc down, and the edge now waits for it

nigrosimone:
The `production-stack` numbers saved with #1117 come from a run where `authsvc` was not there, so they say nothing about the entry. In that run the edge answered 260987 requests with a 5xx, which is 61.2% of them, and the read mix asks for `/api/*` on 12 of its 20 lines, 60%: every request that needed the auth subrequest failed and the server was never reached. The saved `cpu_breakdown` says the same, `authsvc` is absent and `server` sits at 35% where it was 631% before.

The four runs before it, on the same stack:

| run | rps | 5xx | cpu_breakdown |
| --- | ---: | ---: | --- |
| #1100 | 59123 | 0 | edge 2991% \| authsvc 498% \| server 569% |
| #1101 | 58327 | 0 | edge 2925% \| authsvc 444% \| server 574% |
| #1104 | 58753 | 0 | edge 2969% \| authsvc 461% \| server 576% |
| #1106 | 58573 | 0 | edge 2965% \| authsvc 499% \| server 631% |
| #1117 | 32959 | 260987 | edge 3000% \| server 35% |

Nothing in #1117 touched the stack, it changed one handler and the version, so this is the run and not the entry.

The change here is the guard: `authsvc` gets a healthcheck on its own `/_health` and the edge waits for `service_healthy` instead of `service_started`, so a stack that comes up without it fails to start rather than measuring 5xx at full speed. The probe runs every second only while the stack is coming up, then every 30s, so it costs the measurement nothing.

Please rerun `production-stack` and save it, the board is showing that run as a 44% drop.

---

One more thing in this PR, and it does touch the entry: the json route and the two crud writes now use `res.json(obj)` instead of `type('application/json').send(JSON.stringify(obj))`. Same bytes on the wire, same content-type and same `server` header, and 0.9 to 1.2 us less per response on the 50 item payload, as medians over three processes with two identical control arms 200 to 400 ns apart. `res.json` writes the content-type straight into the header object, where `type()` goes through `set()` and pays a lowercase, a charset regex and a header validation.

So the rerun asked for above will measure that too. If you would rather keep the two apart, say so and I move it to its own PR.


nigrosimone:
C:/Program Files/Git/benchmark -f fulmine -t production-stack --save

Kaliumhexacyanoferrat:
@nigrosimone note that I broke gateway validation today and will need a few changes to bring it back. benchmarks should work anyway.

MDA2AV (approved):

PR #1128, opened 2026-08-14
[WTX] Add support for TLS streams (gRPC)

c410-f3r:
## Description

Let's see if #1123 will help with #993.


c410-f3r:
Validation script states that both unary and stream endpoints are failing but again such a thing does not happen locally. 

Capture d’écran du 2026-08-14 15-33-29

Let's see the benchmarking host.

/benchmark -f wtx-grpc-tls --save

c410-f3r:
Yeap, it seems that nothing changed since #993

PR #1133, opened 2026-08-15
Add fiber framework

nigrosimone:
Adds Fiber (Go) with the profiles it can run: baseline, pipelined, limited-conn, json, json-comp, upload.

Routing and path/query access through the Fiber API, JSON through c.JSON, compression through the fiber compress middleware, body limit raised to 25 MB for the upload profile.

Checked locally by building the image and calling every endpoint of those profiles: query and body sums, per request totals with the multiplier, Accept-Encoding negotiation and the 20 MB upload.

I am not a maintainer of Fiber and I am not involved in its development, I used Claude Code to help me implement the entry correctly.


MDA2AV (approved):

ReneWerner87:
@nigrosimone thx
prefork is disabled by default; we should enable it, since it's also enabled in the [Go FastHTTP](https://github.com/MDA2AV/HttpArena/blob/9d1e55be8f6cec89de09034d024f00f251b65cbe/frameworks/go-fasthttp/main.go#L617-L651) benchmark
https://docs.gofiber.io/api/fiber#enableprefork

PR #1143, opened 2026-08-15
Add vapor framework

nigrosimone:
Adds Vapor (Swift) with the profiles it can run: baseline, pipelined, limited-conn, json, json-comp, upload.

Vapor routes on SwiftNIO, Codable JSON responses, gzip through the server responseCompression setting, streaming body for the upload profile.

Checked locally by building the image and calling every endpoint of those profiles: query and body sums, per request totals with the multiplier, Accept-Encoding negotiation and the 20 MB upload.

I am not a maintainer of Vapor and I am not involved in its development, I used Claude Code to help me implement the entry correctly.


nigrosimone:
What problem has this pr?

MDA2AV:
validate failed, Ill trigger it again

nigrosimone:
The validate failure is not the entry: the 300s watchdog in scripts/validate.sh starts before the docker build, and a from-source Vapor build cannot fit in it. Measured locally: 11m12s for the image on 8 cores, 474s of that is swift build alone, because the dependency graph compiles two vendored BoringSSL copies (swift-crypto and swift-nio-ssl) plus the NIO to Vapor module chain. Vapor has no official prebuilt base image to lean on, the way the userver and drogon entries do.

The container itself is fine: built from this branch, every subscribed profile passes the checks validate.sh makes (baseline GET, POST and chunked, pipeline, json totals, json-comp negotiation, the upload sizes, the raw socket cases).

A possible fix: start the watchdog after the build, so the 300s guards what has to be fast, container start and the endpoint checks, while a hung build stays bounded by the job timeout-minutes. A smaller one: raise VALIDATE_TIMEOUT for the validate job to 900.

PR #1162, opened 2026-08-15
frameworks/node-h3: add benchmark implementation

nigrosimone:
Adds h3, the unjs HTTP framework that Nitro and Nuxt are built on. Note the existing `sark-h3` is an unrelated project's HTTP/3 variant.

One thing to decide, and I went the way that matches what users get: this pins h3 2.x, which is what npm's `latest` tag serves today (`2.0.1-rc.26`), so it is what `npm i h3` installs. It is still a release candidate. The stable line is `1.15.11` under the `1x` tag, and its API is different enough that the entry would have to be rewritten. Say the word and I will send the 1.x version instead.

Written with Claude Opus 5.


MDA2AV:
can you rename this to node-h3 or something? just h3 can be conflicting

nigrosimone:
Renamed to `node-h3`, good catch: `h3` is also the ALPN name for HTTP/3, and the board already has `sark-h3`, `zix-http3` and the `gateway-h3` profile.

Directory, display_name, README and package name all follow. Nothing else changed.

MDA2AV (approved):

PR #1163, opened 2026-08-15
frameworks/bun: add benchmark implementation

nigrosimone:
Adds Bun's own HTTP server, `Bun.serve`, with nothing on top.

Bun is on the board only as a WebSocket echo server, so there is no plain HTTP number for the runtime itself. That number is the floor every Bun framework entry should be read against, hono-bun included.

Six profiles: baseline, pipelined, limited-conn, json, json-comp, upload. Routing is by hand on the URL. Scaling is `reusePort` with one process per core, the way `bun-websocket` does it, not node's cluster module.

Written with Claude Opus 5.


MDA2AV:
same as deno, an engine

nigrosimone:
Done, `type` is now `engine`, same as deno. `Bun.serve` is the same shape: you build the Response and its headers yourself.

Pushed on top of the saved benchmark commit, nothing else changed.

MDA2AV (approved):

PR #1164, opened 2026-08-15
frameworks/deno: add benchmark implementation

nigrosimone:
Adds Deno's own HTTP server, `Deno.serve`, with nothing on top. Same reason as the Bun one: Deno is on the board only as a WebSocket echo server, so the runtime has no plain HTTP baseline.

Six profiles: baseline, pipelined, limited-conn, json, json-comp, upload. Routing is by hand on the URL. Scaling is `reusePort` with one process per core, the way `deno-websocket` does it.

Written with Claude Opus 5.


MDA2AV:
This should be put in engine category, does not properly handle http protocol, requires headers to be defined in the implementation.

nigrosimone:
Done, `type` is now `engine`. Agreed: `Deno.serve` hands you a Request and takes a Response you build yourself, headers included, so it belongs with hyper and bjoern rather than with the frameworks.

Pushed on top of the saved benchmark commit, nothing else changed.

MDA2AV (approved):

PR #1165, opened 2026-08-15
frameworks/go-stdlib: add benchmark implementation

nigrosimone:
Adds the Go standard library `net/http` with the pattern routing of Go 1.22, no framework on top.

Go has six entries today and none of them is the standard library, which is where every Go service starts and what the framework entries should be read against.

`mode` is `tuned`: the standard library has no response compression, so json-comp calls `compress/gzip` by hand.

Written with Claude Opus 5.


MDA2AV:
I see that on this and other go entries we need to set headers like  application/json manually, is this a Go thing?

nigrosimone:
Yes, but it is the standard library rather than the language.

`net/http` picks the Content-Type by sniffing the first bytes of the body, `http.DetectContentType`, which follows the mimesniff spec. That algorithm has no rule for JSON, so JSON comes out as `text/plain; charset=utf-8` unless you set the header yourself. Checked on go 1.25:

```
no explicit header -> Content-Type: text/plain; charset=utf-8
explicit           -> Content-Type: application/json
```

The Go frameworks on the board do not need it because they ship a JSON helper that sets it: gin and echo use `c.JSON`, fiber uses `c.JSON`. The ones without such a helper set it by hand, which is what the existing `go-fasthttp` entry already does with `ctx.SetContentType("application/json")`.

So this entry and chi set it manually for the same reason: the standard library has no JSON response helper, and chi is a router rather than a framework, so it adds none.

MDA2AV (approved):

PR #1174, opened 2026-08-15
frameworks/kemal: add benchmark implementation

nigrosimone:
Adds Kemal, and with it Crystal, which is not on the board at all.

Written with Claude Opus 5.

### DIFF
--- a/frameworks/kemal/Dockerfile
+++ b/frameworks/kemal/Dockerfile
@@ -1,0 +1,16 @@
+FROM crystallang/crystal:1.21.0 AS build
+WORKDIR /app
+COPY shard.yml shard.lock ./
+RUN shards install --production --frozen
+COPY src ./src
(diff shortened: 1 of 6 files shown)


nigrosimone:
The failed run is the build on the benchmark host, not the entry logic: shards install crashes the crystal compiler with "Unhandled exception: Arithmetic overflow (OverflowError)". The same build passes the validate runner, which has 4 CPUs, and the benchmark host reports 128 available CPUs, exactly one past what an Int8 holds, so a cpu count kept in a too-small integer somewhere in the toolchain is the suspect.

A workaround worth trying in the Dockerfile is bounding the affinity of the two build steps, taskset -c 0-63 for both shards install and crystal build. If the built server reads the cpu count at runtime the same way, it may need the same cap when it starts.

PR #1176, opened 2026-08-15
frameworks/warp: add benchmark implementation

nigrosimone:
Adds Warp with WAI, and with it Haskell, which is not on the board at all. Warp has a long history near the top of TechEmpower, so its absence here was conspicuous.

Written with Claude Opus 5.


MDA2AV:
can we ping or somehow let these frameworks maintainers that we are adding their framework?

MDA2AV (approved):

PR #1182, opened 2026-08-16
frameworks/fulmine: bump to fulmine.js 5.12.2

nigrosimone:
fulmine.js 5.12.0 is out, this bumps the entry to it.

The change that matters for the profiles is on the body path: a chunked request body is now collected natively in one callback, where before it was one callback and one copy per chunk. The rest of the release is router and settings work on paths the entry already takes, plus fixes.

No changes to the entry itself.


nigrosimone:
The three compose FAILs are not this bump: the gateway and production-stack compose files pin cpuset 0-15,64-79 and 16-31,80-95, which are the benchmark host cores, and the validate runner has 0-3. Nothing sets PROXY_CPUSET or SERVER_CPUSET during validation, so the defaults apply and docker refuses to create the container. It is the same story as REDIS_CPUSET, which validate.sh already leaves unpinned because validation is correctness-only; the compose profiles need the same treatment. Any gateway-subscribed entry has this on a small runner, trillium carries the same defaults.

nigrosimone:
echo-ws, echo-ws-pipeline, echo-ws-limited and production-stack, are broken!?

MDA2AV:
> echo-ws, echo-ws-pipeline, echo-ws-limited and production-stack, are broken!?

Hmm looks like something broke prod stack test

` Container httparena-fulmine-production-stack-server-1 Started 
 Container httparena-fulmine-production-stack-authsvc-1 Waiting 
 Container httparena-fulmine-production-stack-authsvc-1 Error dependency authsvc failed to start
dependency failed to start: container httparena-fulmine-production-stack-authsvc-1 exited (101)
[FAIL] gateway compose up failed
`

nigrosimone:
Bumped to 5.12.2. 5.12.1 and 5.12.2 are correctness fixes only, all found by fuzzing the router against Express: request framing (a repeated or unparseable `Content-Length` is refused instead of routed), `router.param()` callbacks running for parameters the router did not capture, an error under OPTIONS not reaching the error handlers, HEAD not entering a matched layer, and the overlap analysis reading a wildcard or an optional group as literal text.

None of it touches the hot path. Measured with the repo's own A/B and a null control from the same sitting: hello-world 1.5% and routes-1000 1.9%, against a null floor of 2.0%, so nothing moved that the harness can see. The number of routes still registered natively is unchanged on every benchmark scenario, which is the thing that change could have hurt.

Worth flagging about the previous run: 10 of the 38 tests this entry subscribes to produced no result. The two `production-stack` rows are missing because the gateway compose failed to come up, with the shared `frameworks/_shared/authsvc` container exiting 101 while the edge waited for it to be healthy, and the 8 `echo-ws*` rows are absent with nothing in the visible log to say why. The authsvc image had built fine, so it looks like a startup problem rather than the entry. Not fulmine-specific either, that sidecar is shared with `aspnet-minimal_nginx`.

/benchmark --save

nigrosimone:
👍 From our side this is ready: the entry itself is unchanged, only the dependency moved, and 5.12.2 is on npm. Mergeable as far as we are concerned.

Left as a draft only until the benchmark run above comes back, so the saved numbers match the version being pinned. Happy for it to be merged whenever suits you.

MDA2AV (approved):

PR #1194, opened 2026-08-16
site: gold, silver and bronze medals on the board and every framework page

nigrosimone:
## Description

The board publishes a rank for every entry, but nothing says "this one leads a category". This adds achievements: gold, silver and bronze for a top three place, on three axes - the composite of a family, the same composite among entries written in the same language, and each single test profile.

Nothing is scored again. Family and language medals are read out of the badge index, profile medals use the same `eff()` and the same scored connection set the composite uses, so a medal, a badge and the table below it always agree. Every medal is taken in the entry's own league, and only on profiles that entry is scored on. A field needs three entries and somebody has to be beaten, so bronze needs four and last place is never a medal.

The board shows the counts next to the name, with the full list in the panel that opens on a row click. Every framework page gets an Achievements table, and there is a new Knowledge Base page under Scoring.

`badge_composite()` was refactored to share `eff()` and the competition ranking with the new code; the badge index it writes is byte-identical and badge parity is 776/776.

image
image
image

### DIFF
--- a/scripts/gen_leaderboard_data.py
+++ b/scripts/gen_leaderboard_data.py
@@ -1168,6 +1168,66 @@ def badge_aggregate(profiles, results):
     return {"avg": avg, "mem": amem, "bw": abw, "tpl": atpl}
 
 
+def _scored_for(prof, meta, pid, fw):
+    """scoredForType() in index.html. infraScored is read before the `scored`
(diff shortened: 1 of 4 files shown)


MDA2AV:
We had a similar feature in the old site, but used bananas haha

MDA2AV (approved):

PR #1195, opened 2026-08-16
site: /match/, a weighted matcher that ranks entries by what the reader needs

nigrosimone:
## Description

The board answers "who is fastest overall". It does not answer "what should I use", which depends on what you are building: a JSON API on two cores and a WebSocket fan-out do not want the same entry. `/match/` lets the reader weigh 17 dimensions from 0 to 5 - protocol families, workloads, and cost in memory, CPU per request and p99 - and ranks every entry against those weights, with the reason it placed shown on each row.

Nothing new is measured. Family dimensions are the composite as a percentage of the field leader, workload dimensions are the mean normalized RPS over the profiles they cover, and costs are best-in-field over own value. Everything is computed inside the entry's own league, so the page makes the league a choice and not a filter - a 100 among frameworks and a 100 among reverse proxies are two different competitions. Language and tuned are filters on top.

The page carries its own data (~130 KB) and does not load `data.js`. Weights live in the URL, so a set of weights is a link. There is a Knowledge Base page under Scoring explaining the formula.

`badge_composite()` was refactored to share `eff()` and the scored-profile test with the new code; the badge index it writes is byte-identical and badge parity is 776/776.

image
image
image


MDA2AV:
Looks quite nice!

Just a few notes, the "COST" part may not be a good idea, latency values are not relevant and may lead to people confuse with the actual RTT latency which is not related with this. The CPU per request doesn't make much sense because if a framework exhibits lower performance is because it doesn't scale properly with high core/thread count, not a good thing.

nigrosimone:
About COST. Should I remove it, or do we want to change the label name? 
Perhaps RESOURCE?

MDA2AV:
> About COST. Should I remove it, or do we want to change the label name? Perhaps RESOURCE?

we can keep but the "Low CPU per request" should be removed, it is misleading because a framework with low CPU usage is likely poorly optimized to high cpu core count

Kaliumhexacyanoferrat:
While this is a fun feature, I probably would not use this to decide on a framework as it only cares about performance. For a real comparison there would be stuff like popularity, activity, certain features (e.g. JWT auth) etc. - sounds like another Major project tbh.

nigrosimone:
> While this is a fun feature, I probably would not use this to decide on a framework as it only cares about performance. For a real comparison there would be stuff like popularity, activity, certain features (e.g. JWT auth) etc. - sounds like another Major project tbh.

Closed. Agree this Is a separate project with blackjack and hookers

image

PR #1199, opened 2026-08-17
Varnish polish

gquintard:
take a page out of #1197 and disable logs.

Also, lowercase `display_name` to align to the repository style.

## Description

---

| Command | Description |
|---------|-------------|

```
/benchmark -f genhttp-11 --compare genhttp
```

---

<details>

```bash
```

</details>


gquintard:
@MDA2AV, this looks indeed better, but I think the benchmark setup is quite noisy. Getting -4.2% while doing less work is quite weird.

MDA2AV:
> @MDA2AV, this looks indeed better, but I think the benchmark setup is quite noisy. Getting -4.2% while doing less work is quite weird.

I see, there is a typical +-2-3% expected noise with these 5 second runs, pipelined being the noisiest of all, that is about as good as it gets with our current setup budget. Our "pilot" entry does hit 1-2% noise consistently with daily runs done over 60 days but it's a very stable C implementation with less than 10k lines of code running on a custom io_uring transport with fixed parameters all the way. 

I will trigger some more runs here so we can measure how noisy it is, server is also quite hot today because it is close to 40ºC outside here in Portugal 

Also I'd expect higher swings in the pipelined test, it is very difficult for that test to be stable and to be honest, it is not much of a relevant one. On techempower many entries had 10% perf swings between runs in pipelined and they had some very stable harness

To increase stability we'd need to bump each run duration to 15-30 seconds at least and do best of 5 or an average, unfortunately as of today, I cannot afford increasing that because our full runs would take too long and consume too much energy, in the long run that will be improved

gquintard:
Gotcha, thank you for the context. I just wanted to make sure you were aware. I'm very grateful for your work here, and I'll take noisy benchmarks over none :-)

MDA2AV:
Github actions are going through some downtime so I'll bench later.

There shouldn't be a lot of noise though, at least acceptable for the type of benchmarks we cover, we also have this https://timeline.http-arena.com/#test=baseline-4096 it daily reads the current data from the site,  currently we are executing bench runs on demand whenever a pr opens, if you want to measure how noisy it is you can set up a chore or open a pr everyday for a week or something and see the deltas there

gquintard:
Oh, very nice, thank you!

MDA2AV (approved):

PR #1204, opened 2026-08-19
[WIP] Update Ktor with 3.6.0

bjhham:
## Description

In 3.6.0, we're introducing HTTP/3 support and improving memory allocations in a few places.  In anticipation of the release, I thought I'd run an early access build against the HttpArena benchmarks to see how it performs.  I'll update this PR with the proper release artifact after it is published.

## Fixes

* See https://github.com/ktorio/ktor/pull/5822 - our initial implementation was getting pinned to a single thread, so we had to make several changes to the Netty implementation to use multiple ports / load balance over multiple threads.  This should take advantage of parallelism in the benchmarks now.  There were also a few tweaks that should improve overall throughput.

### DIFF
--- a/frameworks/ktor-ghost/Dockerfile
+++ b/frameworks/ktor-ghost/Dockerfile
@@ -9,7 +9,7 @@ RUN gradle buildFatJar --no-daemon -q
 FROM eclipse-temurin:21-jre
 WORKDIR /app
 COPY --from=build /app/build/libs/ktor-ghost-httparena.jar .
-EXPOSE 8080 8443
+EXPOSE 8080 8081 8082 8443
 ENTRYPOINT ["java", \
   "-server", \
   "-XX:+UseG1GC", \
--- a/frameworks/ktor-ghost/build.gradle.kts
+++ b/frameworks/ktor-ghost/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
     alias(libs.plugins.kotlin.jvm)
     alias(libs.plugins.kotlin.serialization)
-    alias(ktorLibs.plugins.ktor)
+    alias(libs.plugins.ktor)
     alias(libs.plugins.kotlin.ksp)
 }
 
--- a/frameworks/ktor-ghost/gradle/libs.versions.toml
+++ b/frameworks/ktor-ghost/gradle/libs.versions.toml
@@ -18,6 +18,7 @@ postgresql = { module = "org.postgresql:postgresql", version = "42.7.4" }
 hikaricp = { module = "com.zaxxer:HikariCP", version = "6.2.1" }
 
 [plugins]
+ktor = { id = "io.ktor.plugin", version = "3.5.2" }
 kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
 kotlin-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", version.ref = "kotlin" }
 kotlin-ksp = { id = "com.google.devtools.ksp", version = "2.3.9" }
--- a/frameworks/ktor-ghost/meta.json
+++ b/frameworks/ktor-ghost/meta.json
@@ -14,8 +14,11 @@
   "repo": "https://github.com/ktorio/ktor",
   "enabled": true,
   "tests": [
-    "baseline", "latency-1m", "latency-10k",
+    "baseline",
+    "latency-1m",
+    "latency-10k",
     "limited-conn",
+    "async",
     "pipelined",
     "json-comp",
     "json-tls",
@@ -29,7 +32,8 @@
     "baseline-h3",
     "static-h3",
     "echo-ws",
-    "echo-ws-pipeline"],
+    "echo-ws-pipeline"
+  ],
   "maintainers": [
     "bjhham",
     "juanchurtado1991"
--- a/frameworks/ktor-ghost/meta.json
+++ b/frameworks/ktor-ghost/meta.json
@@ -26,6 +26,8 @@
     "static-h2",
     "baseline-h2c",
     "json-h2c",
+    "baseline-h3",
+    "static-h3",
     "echo-ws",
     "echo-ws-pipeline"],
   "maintainers": [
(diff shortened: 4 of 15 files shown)


bjhham:
The h3 bottleneck looks better now except for static files going down to 37 rps lol.  I'll investigate this one next.

bjhham:
I think I solved this bottleneck in the pipelining, though it doesn't manifest in benchmark-lite.  I'll need another run on this updated dev build to see if it works 🤞

MDA2AV:
Hmm the symptom is websockets at high connection (16k), this can be annoying to find the issue. Can you reproduce 16k ws echo locally? If you have a ktor build with a playground or example for this I can debug directly on the server

bjhham:
I'll investigate the WS issue later.  It looks like it could be some regression from pending fixes around connection cleanup.

The pipelined profile still being lower than the baseline looks like there's another bottleneck in pipelining with many connections too...

bjhham:
Added consolidated flush calls and played around with the threads.  This helped with the local benchmarks anyway, we'll see how it goes on the server 🤞

MDA2AV:
Just a quick note, currently baseline test is being reworked because its results swing a lot between runs, this is related with the fact that some frameworks parse CL much better than TE-chunked and the request distribution varies between runs. Also we are moving away from "all out" tests that try to push as much throughtput as possible in a time frame, to instead pinning the throughput and measure CPU efficiency and tail latencies that spike with garbage collected languages, this typically yields a much more interesting analysis as we can see close to idle and typical workload metrics.

MDA2AV:
```
==============================================
=== ktor / echo-ws / 16384c (tool=gcannon) ===
==============================================
[info] waiting for server...
[info] server ready

[run 1/3]
gcannon v0.5.3 [WS]
  Target:    localhost:8080/ws
  Threads:   64
  Conns:     16384 (256/thread)
  Pipeline:  1
  Req/conn:  unlimited (keep-alive)
  Expected:  200
  Duration:  5s


  Thread Stats   Avg      p50      p90      p99    p99.9
    Latency    2.52s    2.47s    3.42s    4.64s    4.75s

  11314 frames sent in 5.00s, 506 frames received
  Throughput: 101 req/s
  Bandwidth:  272.90KB/s
  WS upgrades: 10808
  WS frames:   506
  Latency samples: 506 / 506 responses (100.0%)
[info] CPU 1302.4% | Mem 3.2GiB

[run 2/3]
gcannon v0.5.3 [WS]
  Target:    localhost:8080/ws
  Threads:   64
  Conns:     16384 (256/thread)
  Pipeline:  1
  Req/conn:  unlimited (keep-alive)
  Expected:  200
  Duration:  5s


  Thread Stats   Avg      p50      p90      p99    p99.9
    Latency   891.87ms   204.30ms    3.06s    4.85s    4.97s

  23647 frames sent in 5.00s, 8960 frames received
  Throughput: 1.79K req/s
  Bandwidth:  382.16KB/s
  WS upgrades: 14907
  WS frames:   8960
  Latency samples: 8960 / 8960 responses (100.0%)
[info] CPU 88.7% | Mem 5.3GiB

[run 3/3]
gcannon v0.5.3 [WS]
  Target:    localhost:8080/ws
  Threads:   64
  Conns:     16384 (256/thread)
  Pipeline:  1
  Req/conn:  unlimited (keep-alive)
  Expected:  200
  Duration:  5s


  Thread Stats   Avg      p50      p90      p99    p99.9
    Latency   635.10ms   78.60ms    2.35s    4.77s    4.93s

  29846 frames sent in 5.00s, 14978 frames received
  Throughput: 2.99K req/s
  Bandwidth:  394.94KB/s
  WS upgrades: 15047
  WS frames:   14978
  Latency samples: 14978 / 14978 responses (100.0%)
[info] CPU 193.1% | Mem 5.3GiB
```

for 16k connections, we cant reach the 16k upgrades, this could be that maybe the upgrades take very long to process in the 5 second window?

MDA2AV:
would the docker logs help here?

MDA2AV:
if you add the test "echo-ws-limited" this is a upgrade spam test, we can see at low connections, high upgrade count

bjhham:
I think maybe the 16384 ws connection test might've started connecting faster seeing as how the -40% RPS happened from an unrelated change.  I think maybe Ktor just needs to limit TCP connections so that it has a chance to process the requests for this test.

bjhham:
I've tried using 16384 ws connections in benchmark-lite.sh and found that it always gets 0 regardless of the engine, unsure why that is.

Anyway, let's see if connection limits can provide a means for the server from getting overloaded.  This should make a nice safety feature anyway.

MDA2AV:
Teamviewer
Id: 126462749
Pw: yzgyca6j

you can connect from browser with a free acc

if you want to try running on server directly, I installed x11 on the ubuntu server, there is a tar with intellij

MDA2AV:
> I've tried using 16384 ws connections in benchmark-lite.sh and found that it always gets 0 regardless of the engine, unsure why that is.
> 
> Anyway, let's see if connection limits can provide a means for the server from getting overloaded. This should make a nice safety feature anyway.

for that conn level you need to use benchmark.sh, pass LOADGEN_DOCKER=true so it uses the dockerized load gen.
so for example go to project root and execute
LOADGEN_DOCKER=true ./scripts/benchmark.sh ktor echo-ws

bjhham:
I'll need an address for that teamviewer login @MDA2AV 😄

MDA2AV:
> I'll need an address for that teamviewer login @MDA2AV 😄

Hmm all you need is id and pass from TeamViewer site,  I tried it. But maybe I'm missing something, haven't used it much.

I went teamview site, logged with google, put id and pw and got access right away

bjhham:
Here? https://web.teamviewer.com/remote-support
I've never used TeamViewer... It says it can't connect.

MDA2AV:
> Here? https://web.teamviewer.com/remote-support
I've never used TeamViewer... It says it can't connect.


thats the correct link but the teamviewer session in the server dies after a few hours, Ill renew it today later and let you know

MDA2AV:
I added a discord section for server direct access too, if/when you need, just ping there and when I have server access, Ill just dm the details. The server has a running max plan claude cli running too if needed

bjhham:
Websocket issue should be fixed now.  I'll give it another go to ensure nothing else is effected now.

bjhham:
lgtm :shipit: 

I'll close this branch and prepare a fresh and tidy one after we've released 3.6.0.  Thanks @MDA2AV for facilitating this lovely benchmark suite!

PR #1206, opened 2026-08-19
fulmine: bump to 5.15.1

nigrosimone:
Bumps the fulmine entry to 5.15.1. The release is three performance batches over the request, the router and the middleware layer, plus two routing fixes found while testing them.

Where the work went, measured locally with the profiler (CPU per request, against the previous release, with a null control from the same sitting):

- the generic route scan now asks an index for its literal routes instead of walking the table, -67% CPU on a scan-heavy request
- a chain of middlewares pays less per hop, -20% CPU
- a stack of helmet/morgan/cors/cookie-parser, -11% CPU
- the second read of req.query replays the first parse instead of re-scanning

Our own CI benchmark against Express moved the same way on the rows that were not already at the load generator's ceiling.

Nothing about the entry changes except the version, so this is only worth a run to see whether any of it shows on the arena's profiles.


nigrosimone:
C:/Program Files/Git/benchmark --save

MDA2AV (approved):

PR #1207, opened 2026-08-19
fulmine-tuned: the same entry with the two caches a standard entry may not use

nigrosimone:
A second entry for fulmine.js, in tuned mode, which differs from the standard one in three settings and nothing else. Diffing the two `app.js` shows the whole change: same routes, same middleware, same `express.compression()`, same `express.static()`.

| Setting | Standard entry | Here | Why standard may not have it |
|---|---|---|---|
| `file cache` | `false` | `true` | the framework's own default; answers a static request from memory, which standard mode forbids |
| `stat cache` | unset | `1s` | without it the cache still stats the file once per request |
| `connection headers` | `true` | `false` | drops `Connection` and `Keep-Alive` from every response, which express sends |

The first two are the reason the entry exists: they are the only shipped settings a standard entry is not allowed to use, and both are about the static profiles.

The compression level deliberately stays where the standard entry has it, brotli q3. `json-comp` squares the byte ratio, which makes a higher quality look free; measured on this profile's own payload it is not. Quality 4 costs 88% more encode time for 0.2% fewer bytes, quality 5 costs ten times the time for 8%, and against the squared term every quality above 3 still loses, by 13% at q4 and 95% at q11. q3 is the maximum of that function.

The entry subscribes to ten profiles rather than the standard entry's twenty: the ones where these settings can show, plus the baselines to catch a regression. No compose files, so it adds nothing to the gateway or production-stack runs.


MDA2AV:
caching static files in memory is against the rules, even for tuned. That rule was changed, some frameworks still break it but are being looked into. This test  measures file  I/O

nigrosimone:
Understood, and thank you for catching it before the run. I have pushed the fix: `file cache` is now off in this entry exactly as it is in the standard one, so no static file body is held in memory and every request reads the file it answers with.

What is left is `stat cache`, which caches the size and mtime only, never a body. The read still happens per request; what it saves is the syscall that asks whether the file changed, the same trade nginx's `open_file_cache` makes with its `valid` window.

If that is over the line too for the static profile, say so and I will drop it, which would leave this entry differing from the standard one only by `connection headers: false`. At that point it is probably not worth a second entry at all and I would rather close this than take up benchmark time for two headers.

nigrosimone:
One thing worth fixing while you look into the frameworks that still break the rule: the docs are what sent me the wrong way, so anyone else reading them will land in the same place.

`site/content/docs/add-framework/implementation-rules/frameworks/tuned.md` line 15 still lists it under **What is allowed**:

```
- Memory-mapped files and in-memory static file caching
```

The standard page is unambiguous by contrast, "Static files must be read from disk ... No in-memory caching, no memory-mapped files, no pre-loaded file buffers", and reading the two side by side the natural conclusion is that the restriction is exactly what standard mode adds and tuned mode lifts. That is how I built this entry.

Happy to send the doc change as its own PR if you tell me how you want it worded, since the rule now applies to every mode rather than being a tuned exception.

MDA2AV:
https://www.http-arena.com/docs/add-framework/implementation-rules/frameworks/tuned/

that was updated like 30m ago

nigrosimone:
Thanks for updating it, and the new wording answers my question exactly: the line draws at file **bodies**, "Every entry reads static file bodies from disk on each request", with pre-rendered response headers still fine.

This entry is on the right side of that as pushed. `file cache` is off, so no body is held anywhere and every request reads the file it answers with. What it keeps is `stat cache`, which holds size and mtime only, the two values that build content-length and last-modified, and never touches the body.

Re-running the request below, since the head moved after the last acknowledgement.

MDA2AV (approved):

nigrosimone:
The site has not picked these results up, and it is not going to on its own. The squashed merge carries the results bot's `[skip ci]` in its message:

```
* Benchmark results: fulmine-tuned  [skip ci]
```

Actions honours that for the whole push, so nothing ran for `deaee6f4`, Deploy Site included. Its last run is `307316c5` at 18:00, two merges ago. The data itself is fine, `site/data/results/fulmine-tuned.json` landed with the merge; it is only the build that never happened.

A `workflow_dispatch` of Deploy Site is all it needs. I cannot fire it myself, it wants admin.

Worth knowing beyond this PR: every entry merged with saved results has the same shape, so the board silently lags by one deploy each time until some later merge touches `site/**` or `results/**` and rebuilds it. Stripping `[skip ci]` from the squash message when merging, or giving the deploy a `schedule:`, would close it for good.

PR #1211, opened 2026-08-20
fulmine: answer gzip on json-comp, both entries

nigrosimone:
Both entries move to fulmine 5.16.0 and answer gzip level 1 on json-comp instead of brotli q3, through the middleware's new `encodings` option.

The host's own numbers say brotli costs about twice the CPU per response here: the 5.15.1 run measured 136us added per request against 71us for the gzip entries, and the squared byte term does not buy that back. Projected on those numbers, gzip scores ~15% higher on all three json-comp rows.


nigrosimone:
/benchmark-multiple --save

(supersedes the request above, which would have run only the first entry; this one runs both)

nigrosimone:
I thought it was a good idea, but the numbers show otherwise.

PR #1226, opened 2026-08-21
completeness: not on the WebSocket and gRPC boards

MDA2AV:
The four completeness axes — routing, middleware, request, response — are the path from an arriving **HTTP** request to a finished response. Two families do not measure that path, so scaling their scores by the grade was measuring one thing and charging for another.

- **WebSocket.** No route to match, no body to hand over, no response to build. The handshake happened once; everything the profile times after it is frames on an open socket.
- **gRPC.** A call *does* have all four — it is dispatched, the request arrives deserialized, the response is serialized back — but the stub generated from the `.proto` does them. Grading the framework for those would be crediting protoc.

## The exemption is on the board, not the entry

23 entries run both HTTP and WebSocket or gRPC profiles, so a per-entry rule would have left them scaled on exactly the boards this is about. Scoring the same entry on two boards with the page's own code:

```
swoole   h1    factor=0.85   score=596  raw=701
swoole   ws    factor=1.00   score= 79  raw= 79
```

Same entry. The second board is not measuring the thing it was marked down for.

Those rows now carry no factor and no `(n)` marker at all — the same treatment engine and infrastructure rows already get, for the same reason: the question is out of scope, not unanswered.

## The 20 entries that drop the field

An entry running *only* WebSocket or gRPC profiles has no board on which the factor would ever apply, so its grade was inert. A grade nothing reads is still a claim about an HTTP path those entries do not have, so it comes out of `meta.json`:

`actix-websocket`, `aspnet-grpc`, `aspnet-websocket`, `beskar-websocket`, `bun-websocket`, `deno-websocket`, `dogrider`, `fleck`, `go-websocket`, `grpc-go`, `mq-bridge-websocket`, `node-websocket`, `quarkus-jvm-grpc`, `spring-jvm-grpc`, `swoole-websocket`, `tonic-grpc`, `workerman-websocket`, `wtx-grpc`, `wtx-grpc-tls`, `wtx-ws`

They were graded on a stretched reading of the axes — "does the library decode frames for you" is not the same question as "does the framework hand you a parsed request". Selected by their own `tests` array rather than by published results, so the 11 of them without results yet are covered too.

Declared entries: **134 → 114**. Distribution over what remains:

| Factor | Before | After |
|---|---|---|
| ×1.00 | 96 | 89 |
| ×0.95 | 14 | 8 |
| ×0.90 | 17 | 10 |
| ×0.85 | 4 | 4 |
| ×0.80 | 3 | 3 |

## Verification

Scoring lives in two copies, so `cmpFactor()` and `_cmp_factor()` both take the scope now, and the header tooltips say which case a board is in.

- `check_badge_parity.js` exercises **all six families** and passes at **783 ranks**
- Each board scored with the page's own lifted code:

```
h1    102 entries | rows showing a factor: 102 | deducted: 22
h2     30 entries | rows showing a factor:  30 | deducted:  8
h3     13 entries | rows showing a factor:  13 | deducted:  4
gw      5 entries | rows showing a factor:   5 | deducted:  0
ws     33 entries | rows showing a factor:   0 | deducted:  0
grpc    8 entries | rows showing a factor:   0 | deducted:  0
```

- Parity passing alone would not prove the gate fires (it only proves the two sides agree), hence the per-board check above
- Rendered the WebSocket board in a browser: no `(0)` or `(-n)` on any row
- Docs updated — `completeness.md` gets a "Not on the WebSocket and gRPC boards" section, and `meta-json.md` says to omit the field on ws/grpc-only entries


joanhey:
It's the more bizarre change that I see in a test or benchmark !!!!

We are closing to think different.
PHP never need a routing, as is the same url to a path. The middleware is the class that extend it. The request and response are direct.

Almost all languages need a routing, because they can`t connect directly to a url. And the same for request and response.
We need to think different, and open our mind and the people who can think different apart from this logic.

joanhey:
The first think that need to change is the composite score.

Now is not valuable in any way as a performance value.  We need to divide the result with the tests profiles than pass.
Actually any new profile is an advantage to the frameworks than add it.
But not in any way faster than others frameworks.

The composite score need to be:  composite score / profiles tested. So I'll be a performance value.
Any framework that have more tests have more score, even if it's slower.

MDA2AV:
> The first think that need to change is the composite score.
> 
> Now is not valuable in any way as a performance value. We need to divide the result with the tests profiles than pass. Actually any new profile is an advantage to the frameworks than add it. But not in any way faster than others frameworks.
> 
> The composite score need to be: composite score / profiles tested. So I'll be a performance value. Any framework that have more tests have more score, even if it's slower.

This is the most bizarre thing I've read on a benchmark. If a framework can't implement those tests, can it even be called a web framework? Maybe a framework, not web for sure. We don't promote laziness, specially in an era where any AI agent will wire the entire entry for you in minutes lol. 

composite score / profile? So I am telling people to only implement their fastest test?? what?!

MDA2AV:
> It's the more bizarre change that I see in a test or benchmark !!!!
> 
> We are closing to think different. PHP never need a routing, as is the same url to a path. The middleware is the class that extend it. The request and response are direct.
> 
> Almost all languages need a routing, because they can`t connect directly to a url. And the same for request and response. We need to think different, and open our mind and the people who can think different apart from this logic.

https://github.com/MDA2AV/HttpArena/blob/main/frameworks/laravel/routes/api.php
maybe this isnt PHP?

joanhey:
Laravel with Octane is not normal PHP.

Bye !!!

MDA2AV:
```php
$https->onMessage = static function ($connection, $request) {

    if(str_starts_with($request->path(), '/json/')) {
        $count = explode('/', $request->path())[2];
        $m = $request->get('m', 1);
        $total = [];
        $i = 0;
        while ($i < $count) {
            $item = JSON_DATA[$i++];
            $item['total'] = $item['price'] * $item['quantity'] * $m;
            $total[] = $item;
        }
        $connection->headers = ['Content-Type' => 'application/json'];
        return $connection->send(json_encode(['items' => $total, 'count' => $count], JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES));
    }

    return $connection->send(new Response(
        404,
        ['Content-Type' => 'text/plain'],
        '404 Not Found')
    );
};
```

No, you are not telling me this is production code.

joanhey:
Bye !!!

PR #1234, opened 2026-08-21
fulmine-tuned: pipeline the postgres queries with pg-telaio

nigrosimone:
The fulmine-tuned entry routes its postgres queries through [pg-telaio](https://github.com/nigrosimone/pg-telaio), a tagged template over the pg Pool that pipelines the queries on a few shared connections instead of checking out one connection per query. Statements stay parameterized and prepared, and the per-worker connection budget does not grow: perWorker pipelined connections plus one pool connection as overflow, in place of a pool of perWorker.

The library is new, so this stays in the tuned entry only; the standard entry is untouched.

Validated against the unpatched app on the same machine, one worker, same budget: every DB endpoint answers byte-identical status, headers and body, the upsert included. The crud run here measured +6.3% on fulmine-tuned.

One SQL change rides along: the crud insert's ON CONFLICT clause says excluded.name instead of $2 back references, which a template cannot repeat. Same semantics, verified on the resulting rows.


nigrosimone:
C:/Program Files/Git/benchmark-multiple -f fulmine,fulmine-tuned -t async-db

nigrosimone:
C:/Program Files/Git/benchmark-multiple -f fulmine,fulmine-tuned

MDA2AV:
This is only allowed for fulmine-tuned

nigrosimone:
You are right, pg-telaio is a new library and the standard rules ask for an established one. Reverted: frameworks/fulmine is back to main, the change lives in fulmine-tuned only. Title and description updated too.

nigrosimone:
benchmark -f fulmine-tuned

MDA2AV:
Can you check the validation failure? I'll trigger benchmark after

nigrosimone:
About the fortunes XSS failure: I think that step flaked, the same code passed it two hours earlier.

- Run 32507220789 (17:24) validated the same fulmine-tuned files and reported PASS [GET /fortunes XSS escape]. The only commit in between, 2836c66, touches frameworks/fulmine, which is back to main.
- Both runs report the same page: body size=24536B and <tr> count=202. So the rendered output did not change between the two.
- Reproduced here with this repo's own seed (the fortune block of data/pgdb-seed.sql, 200 rows plus the runtime one): the page contains &lt;script&gt; and no raw <script>alert, on the JS client and on pg-native in docker. 40 sequential fetches and fetches under concurrent load all give the same 24536-character page.

I cannot explain what the step saw, so a re-run would tell us. If it fails again I will dig further.

MDA2AV (approved):

PR #1254, opened 2026-08-21
python frameworks: add the missing HTTP/1.1 profiles

MDA2AV:
The Python sweep. Twelve entries gain profiles; three are deliberately left
short and are explained below.

| entry | added | tests |
|---|---|---|
| starlette, litestar, django | static, static-tls, json-tls, async-db, crud, api-4, api-16 | 6 → 13 |
| aiohttp | same seven | 6 → 13 |
| sanic | static, async-db, crud, api-4, api-16 | 6 → 11 |
| blackbull | async-db, api-4, api-16 | 12 → 15 |
| flask | static-tls, crud | 11 → 13 |
| bottle | json-comp, crud | 11 → 13 |
| fastapi, uvicorn, fastpysgi-wsgi, mq-bridge-py | crud | +1 each |

`pyronova` was already complete.

### How the pieces were done

**json-tls / static-tls** come from the launcher `fastapi` already used — uvicorn
spawned twice, the second on port+1 with the cert and key. litestar and django
ran uvicorn straight from `CMD` and now share it. aiohttp uses a second
`TCPSite` on the same runner, bound by every forked worker.

**static** bodies are read from disk on every request: Starlette's `StaticFiles`,
litestar's static router, `FileResponse` for django and aiohttp, `file_stream`
for sanic. The encoding is left to the compression each entry already does.

**Postgres** is asyncpg (psycopg where the entry already used it), one pool per
worker with `DATABASE_MAX_CONN` divided by the worker count and headroom left
for `superuser_reserved_connections`. **crud** runs cache-aside on Redis, 200ms
TTL, explicit delete on update.

### Pre-existing bugs fixed along the way

- **mq-bridge-py** forks one serving process per core and each opened a pool of
  the *whole* `DATABASE_MAX_CONN`. On the crud profile's 62-CPU cpuset that asks
  for 62 × 256 connections of a Postgres running `max_connections=256` — the
  same bug #1251 fixed on hono-bun, and crud is exactly the profile that would
  have exposed it.
- **flask** answered `/pipeline` and `/baseline11` with `text/html`. Flask
  defaults a bare body to that and neither handler set a content type, so all
  three Content-Type checks failed.

### Left short, on purpose

- **sanic** — no `json-tls`/`static-tls`. It serves one port per `app.run()`; the
  documented multi-port form (repeated `app.prepare()` then `Sanic.serve()`)
  starts without error but 8081 never answers. Its own TLS helper also raises
  `KeyError: 'subjectAltName'` on the arena's self-signed certificate, so a
  pre-built `ssl.SSLContext` is needed to get that far at all.
- **blackbull** — no `crud`. The collection route works, but a handler declared
  with a path parameter does not also receive `receive`/`send`, so
  `/crud/items/{id}` can neither read a PUT body nor write a response.
- **fastpysgi-asgi** — untouched. It fails **86 checks on unmodified main**
  (static, static-tls and TCP fragmentation, all answering empty), so there was
  no clean baseline to add crud against. Needs its own investigation.

### Also worth a look

`uvicorn`, `fastpysgi-asgi` and `fastpysgi-wsgi` all preload static file bodies
into a `STATIC_FILES` dict at startup and serve from memory. The static profiles
forbid that **in every mode**, including `engine`, which all three are. I have
not touched their static path — all three already publish `static` and
`static-tls`, and fixing it would move those numbers — but it is the same
violation class as go-fasthttp in #1253 and swoole in #1244.

### Validation

starlette, litestar, django, aiohttp, fastapi, uvicorn, fastpysgi-wsgi **58/0**;
mq-bridge-py **74/0**; flask, bottle **56/0**; sanic **47/0**; blackbull **66/1**,
where the one failure is `[frag: POST, chunked]` and reproduces identically at
**63/1** on the unmodified entry.

### Still open

`robyn` (7 profiles), `slimeweb` and `bjoern` (3 each), plus the three noted
above.


MDA2AV:
/benchmark-multiple -f aiohttp,sanic,blackbull,flask,bottle,fastapi,uvicorn,fastpysgi-wsgi,mq-bridge-py --save

PR #1255, opened 2026-08-22
fulmine, fulmine-tuned: fulmine.js 5.18.1, gzip on json-comp, pg-telaio 0.1.2

nigrosimone:
pg-telaio 0.1.2 fixes a false positive in its slow-query guard, and the guard is then turned off for this entry: every query these profiles run is a point read of a few milliseconds, so there is no slow query for it to catch. production-stack is 50,177 rps at 1024 connections, +6.9% on top of the +21.8% the earlier run on this branch measured.

Both entries build the json items with a preallocated loop instead of slice().map(), and fortunes pushes the runtime row onto the driver rows and sorts them in place instead of copying them first.

json-comp changes encoding, which reverses what this entry did before, and fulmine.js is why: 5.18.0 gzips a whole body on a stream it keeps rather than on one built and thrown away per call, and a brotli stream cannot be kept that way. Brotli's body is 10% smaller and the score squares that, so this was only worth trying because the call got cheaper. The run says it was: **+43.3%, +43.2% and +38.4% rps** on the three rows, with bytes per request going 1379 to 1506, which is a net gain on the scored figure.

5.18.1 then copies a request body chunk out of a view of the uWS buffer instead of out of a slice of it, which is the same bytes without an ArrayBuffer allocated per chunk. It is not what these rows measure; it is here because the entry should not pin an older version than the one it was built against.


nigrosimone:
Wait for https://github.com/MDA2AV/HttpArena/pull/1255#issuecomment-5379078663

MDA2AV (approved):

PR #1256, opened 2026-08-22
php frameworks: add the missing HTTP/1.1 profiles

MDA2AV:
The PHP sweep. `swoole` and `true-async-server` each gain `crud`, the one profile
they were missing.

| entry | added | tests | validation |
|---|---|---|---|
| swoole | crud | 14 → 15 | 67/0 |
| true-async-server | crud | 19 → 20 | 80/0 |

### Scope

Of the 14 PHP entries, **9 are in scope**. Out: `workerman-websocket` and
`swoole-websocket` (WebSocket-only, two tests each), `workerman` and `ngx-php`
(disabled in #1240 for failing validation), and `php` (already complete).

### The cache is in memory, not Redis

The profile's default backend is an in-process cache; the sidecar exists for
runtimes where a per-process one fragments. Neither of these needs it — but
neither has a heap to share either, and the profile reads ids at random out of
50,000, so a plain array does fragment: the MISS-then-HIT check lands on two
different workers and sees MISS twice.

- **swoole** uses **`Swoole\Table`**, allocated before the workers fork.
- **true-async-server** runs one process with N worker threads, each with an
  isolated PHP context, so it gets `SharedCache`: a fixed slot table backed by
  **a file on `/dev/shm`** — tmpfs, so page-cache speed. Key hashes straight to
  its slot, so it self-evicts and never grows. No lock; a crc turns a torn read
  into a miss, costing one Postgres read.

`SharedCache` was `shmop` first — **80/0 locally on 32 cores, then nothing but
misses on the 4-core CI runner**. SysV shm limits differ per machine and
`shmop_open` fails silently. Worth knowing before reaching for it elsewhere.

This also avoids a packaging mess: **none of the PHP base images ship `redis` or
`apcu`**, and they disagree on how to add one.

### Three bugs found

- **swoole**: `Crud.php` needed `use Swoole\Http\Request/Response` — type hints in
  a separate file resolve against the global namespace, so every crud call raised
  a `TypeError` and returned an empty body.
- **true-async-server**: it opens its PDO pool per worker on first use, and crud
  went through a new `available()` that only read the flag — a request landing on
  a worker that had not served `async-db` yet answered 500.
- **Both**: a single-row `fetch()` leaves the cursor open, and a pooled PDO will
  not return the connection until it is closed. `crudList` drained its rows in a
  loop and kept working; `crudRead`/`crudCreate` fetched once, so a handful of
  single-item reads exhausted the pool and every later query failed. Fixed with
  `closeCursor()`.

`tags` is a JSONB column, so PDO hands it back as text.

### Reverted: the two frankenphp entries

`frankenphp-trueasync` and `symfony-spawn-franken` were in this branch and are
backed out. Both have problems that predate it and do not belong in a profile
sweep:

1. **Neither entry starts.** The upstream
   `trueasync/php-true-async:latest-frankenphp` image ships a frankenphp binary
   linked against libwebp *without the libraries in it* — exit 127 on
   `libwebpdemux.so.2`. Confirmed against the unmodified entries. Installing
   `libwebp7 libwebpmux3 libwebpdemux2` fixes it.
2. **Once started, both fail the exhaustive TCP fragmentation checks** on
   `/baseline11` — frankenphp-trueasync 4 on CI, symfony-spawn-franken 9 on CI
   and 1 locally. Not new: the frag suite landed in #1210, and neither entry has
   been validated since, because #1223 only changed their metadata and CI
   validates only what a PR touches. Fixing (1) is what let CI reach them.

The profile work for frankenphp-trueasync — json-tls, static-tls, crud, a PDO
pool with headroom (it was capped at `DATABASE_MAX_CONN` exactly and answered
`sorry, too many clients already`), and static bodies read from disk rather than
held in memory — is worth keeping, but cannot land while the entry is red for
unrelated reasons.

### Still to come

`slim` and `laravel` (7 profiles each), `symfony-spawn-tas` (4), `mark` and
`hyperf` (3 each).


MDA2AV:
/benchmark-multiple -f swoole,true-async-server -t crud --save

Issue #1275, opened 2026-08-22
Static, Static TLS: Entries do not serve the same amount of data (case in point Bun)


uNetworkingAB:
If you look at fulmine and Bun, fulmine has higher composite score even though Bun beats fulmine in everything except static, static TLS. And then you look into why and you see exactly the same pitfall as the uWS.js entry had:

https://github.com/MDA2AV/HttpArena/blob/652b77856f1b82cb76f7f52d392a15efb44cfbaa/frameworks/bun/server.ts#L143

It does not pick the .bz compressed file version. And so it ends up scoring HALF of what it should.

You probably need to make this a failure so that this is no longer a silent failure that massively penalizes those who do not pick the .bz files. It's not even remotely close to apples-apples if you compare non-compressed file RPS against precompressed file RPS.

uNetworkingAB:
> This entry is `standard`, so the .br/.gz files the
// harness leaves on disk are left alone: hand-rolled suffix lookup is a tuned
// technique and Bun.serve has no documented pre-compressed static API.

This is nonsense. Your whole test basically boils down to:

`those who know the secret trick - to serve compressed files - are guaranteed to win the benchmark because it's entirely impossible to win by sending the full uncompressed file.`

You're not benchmarking servers, you're benchmarking how well the server know your secret trick. Either change the static benchmark to measure GB/sec instead of req/sec or remove those servers who don't know the trick.

It's pointless to participate in a benchmark where the competition is having its perf. halved for bureaucratic reasons.

uNetworkingAB:
> Is this such a complex intricate thing to ask to frameworks

Well, the problem is that you're assuming a filename X with extension .bz is the same contents as the compressed file without that .bz file name, and that servers know this. That's a `decent convention` but it's `not standard`, so essentially no one other than you follow this rule.

Maybe this is well known in .NET but it's still not standard. A file with extension .bz can be entirely different from the same name without it. For instance, maybe the compressed variant is a snapshot from a few versions back? Etc, etc.

Edit: "maybe the compressed variant is a snapshot from a few versions back?" -> this is actually something I do myself often, now that I think of it.

uNetworkingAB:
> What is then a test that suits to Bun, a plain text Hello World?

That's kind of the problem with these tests - they often overengineer their purpose just because (for instance by adding database tests which have nothing with the web server at all to do) something that basically is quite slim:

You can really only reliably test a few basic things:

1. Standards compliance (you have good tests here)
[...]

PR #1277, opened 2026-08-23
let's help bun serve some static files.

MDA2AV:
Closes the gap raised in #1275.

## The problem

The static rule allowed the `.br`/`.gz` variants **only behind a documented framework API**. In practice that made the profile turn on whether a framework happened to ship one feature. An entry sending a **200 KB** body against entries sending the **47 KB** brotli variant of the same file is not doing the same work, and that gap swamped everything else the profile measures.

`bun`, `elysia` and `hono-bun` all served full-size bodies for exactly that reason.

## The rule change

**Selecting an already-compressed file is not compression.** The bytes already exist on disk next to the original, so picking one is a read of a different path. The rule now allows it for every entry type — through the framework's own API where there is one, otherwise off `Accept-Encoding` in the entry.

What stays out is unchanged:
- no compressing at runtime by hand in `standard`
- the cache must be the framework's own, not one assembled in the entry
- it must still follow the disk

Updated in all four places the rule appears (`static`, `static-tls`, `static-h2`, `static-h3`) plus the standard-mode implementation rules.

## The entries

| entry | how |
|---|---|
| `hono-bun` | Hono's own `serveStatic({ precompressed: true })` — it already had the feature, the entry just wasn't using it (it had a hand-rolled `Bun.file` route instead) |
| `bun` | no such API exists on `Bun.serve`, so it selects the variant itself |
| `elysia` | `@elysiajs/static` takes no `precompressed` option — checked against the package rather than assumed. Keeps the plugin for everything else and only intercepts when an encoded variant exists |

`bananabread` and `bun-websocket` are also Bun-based but carry no `static` profile, so they are untouched.

## Verified per entry

Binary files stay untouched — `webp`/`woff2` have no variant on disk, and Hono only tries precompressed for compressible content types.

```
app.js  br     -> Content-Encoding: br    Content-Type: text/javascript
app.js  gzip   -> Content-Encoding: gzip
app.js  identity -> no Content-Encoding
hero.webp br   -> no Content-Encoding     Content-Type: image/webp
body size: 47275 served == 47275 on disk  (raw 204800)
```

The staleness probe still passes on both the plain and TLS listeners, so the encoded path follows the disk too.

| entry | validate |
|---|---|
| bun | 68 / 0 |
| hono-bun | 68 / 0 |
| elysia | 68 / 0 |

(`hono-bun` failed the fortunes XSS check on one run and passed on a re-run; that check is intermittent and this diff does not touch that path.)

These three will need a re-benchmark on the static profiles for the change to show up on the board.


MDA2AV:
Re-benchmarking the three entries on the static profiles so the board picks up the pre-compressed bodies.

/benchmark-multiple -f bun,elysia,hono-bun -t static --save

MDA2AV:
And the TLS listener, which serves the same variants.

/benchmark-multiple -f bun,elysia,hono-bun -t static-tls --save

MDA2AV:
**This needs a re-run of `static` for hono-bun** — the saved numbers measured the uncompressed path.

/benchmark-multiple -f hono-bun -t static --save

PR #1278, opened 2026-08-23
docs: warn that Accept-Encoding carries q-values on the static profiles

MDA2AV:
Follow-up to #1277, and the vehicle for the **`static-tls`** save that could not ride along with it.


MDA2AV:
Saving the TLS listener for the three Bun entries. `static` is already saved and correct on main; this is the one still outstanding.

/benchmark-multiple -f bun,elysia,hono-bun -t static-tls --save

PR #1285, opened 2026-08-23
ci: validate every entry when the validator itself changes

MDA2AV:
Opens a full validation sweep over every enabled entry, and fixes the gap that made one impossible.

## The gap

`validate.yml` triggers on `scripts/validate.sh`, then builds its framework list from changed `frameworks/**` paths:

```
changed=$(git diff --name-only origin/$BASE...HEAD | grep '^frameworks/' | cut -d'/' -f2 | sort -u)
```

A PR that only touches the validator matches the trigger and then validates **nothing** — the list is empty and both jobs are skipped by `if: ... != '[]'`.

That is not hypothetical. It happened when the static staleness probe landed in #1267, and again when it was hardened in #1284. Both changed what every entry is checked against; neither ran a single entry.

## The change

- A validator change (`scripts/validate.sh`, `scripts/validate-ws.py`, or this workflow) now covers **every enabled entry**.
- The workflow file joins the trigger paths, since editing it can change what validation means just as much as editing the script. That also makes this PR self-demonstrating: it validates everything.
- Adds `workflow_dispatch` taking an entry name, a comma-separated list, or `all`, so a sweep can be run on demand without inventing a commit.

## Scale

185 enabled entries — **177** on the matrix (GitHub's cap is 256) and **8** compose entries running sequentially on the self-hosted box.

Timeouts were sized for the handful of entries a normal PR touches, so both are raised:

| job | was | now | why |
|---|---|---|---|
| matrix | 30 min | 60 min | a sweep reaches entries that build from source on a four-core hosted runner — h2o, the LTO Rust entries, Swift |
| self-hosted | 90 min | 300 min | one sequential loop now covers all eight compose entries, not one or two |

## What to expect from this run

This is the first sweep against the hardened static probe, so it is also the blast-radius measurement for #1284. From a 12-entry local sample, **no entry that passed the old probe fails the new one** — every failure below already fails on `main`:

| | old probe | new probes |
|---|---|---|
| bun, caddy, elysia, hono-bun, h2o-mruby, slimeweb, genhttp-11-ioxide | PASS | PASS |
| humming-bird, frankenphp-trueasync, genhttp-11, genhttp-11-kestrel, ioxide | FAIL | FAIL |

Two things that sample already surfaced, both independent of the hardening:

- `genhttp-11` and `genhttp-11-kestrel` serve from GenHTTP's `Assets.From`, which never revalidates — 35s after the file changed on disk they still serve the original bytes. Their sibling `genhttp-11-ioxide` passes because `IoxideFiles.From` re-stats per request. Same framework, different handler.
- `ioxide` is stale on both the identity and variant paths out to 35s, despite `Program.cs:100` describing the plain path as read-per-request.

Expect some failures in this sweep to be infrastructure rather than entries — a four-core hosted runner is not the benchmark box, and a few builds are slow or flaky there. Worth reading the first run as a survey, not a verdict.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -6,6 +6,13 @@ on:
       - 'frameworks/**'
       - 'scripts/validate.sh'
       - 'scripts/validate-ws.py'
+      - '.github/workflows/validate.yml'
+  workflow_dispatch:
+    inputs:
+      framework:
+        description: 'Entry name, a comma-separated list, or "all" for every enabled entry'
+        required: false
+        default: 'all'
 
 jobs:
   detect:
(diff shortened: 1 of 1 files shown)


nigoroll:
I am coming from #1289 and am surprised. Could you please help me understand which validation is failing and where I can find more information about it?

MDA2AV:
> I am coming from #1289 and am surprised. Could you please help me understand which validation is failing and where I can find more information about it?

Of course, you can find details here https://www.http-arena.com/docs/test-profiles/h1/isolated/baseline/validation/ last point "Exhaustive TCP fragmentation" this is a added validation step taken from https://github.com/uNetworking/uWebSockets/blob/master/tests/fragment_test.ts modified to also include query parameters.

Run details on Vinyl Cache here  https://github.com/MDA2AV/HttpArena/actions/runs/32719610781/job/97408097735  from 1904 onwards

nigoroll:
> Of course, you can find details here

Excellent, thank you for your help!

MDA2AV:
> Excellent, thank you for your help!

Sorry for the trouble, we are still not 100% sure if some validation failures are false positives hence the "temporary", as we keep adding more validation points. Would be great if you could look into it.

nigoroll:
> we are still not 100% sure if some validation failures are false positives

This is a great attitude, but in the case of Vinyl Cache I can already confirm that it is real and a regression: https://code.vinyl-cache.org/vinyl-cache/vinyl-cache/issues/4575

I love your project already.

PR #1289, opened 2026-08-24
Temporarily disable entries that fail TCP fragmentation, and drop their results

MDA2AV:
From the full validation sweep in #1285. Nineteen entries fail the exhaustive fragmentation check — nine request shapes split at every byte offset, ~1,000 per entry — disabled, with 18 result files removed.

## Severity is not uniform

| worst shape | entries |
|---|---|
| **100% of offsets** | fishcake, helidon-production, helidon-tuned, ioxide, libreactorng, may-minihttp, minima, nestjs, ringzero, rust-epoll, vinyl-cache, web-framework-csharp |
| 88–91% | slim (91%), dart-io (88%) |
| **1–9%** | hical (9%), effinitive (3%), bjoern (1%), dart-zig (1%), blackbull (1 offset of 140) |

The first fourteen cannot parse a request that arrives in more than one read — a real HTTP correctness defect.

The last five are a different kind of defect: an edge case at one split point rather than a parser that cannot reassemble. `bjoern` and `dart-zig` fail 9 of 1047 offsets — exactly one per shape, so a single consistent boundary condition. `blackbull` fails one offset out of 140 and loses 26 result keys for it.


tomas-langer:
The explanation here is not exactly correct for Helidon. 
There is a bug with `Connection: close`, not with fragmentation of requests. 
It still fails, but because the connection does not close (as expected by the probe). It returns 200 correctly.

Hical61:
Confirmed on hical this one's a real bug, The chunked decoder was tracking the wrong offset across fragmented reads (phr_decode_chunked decodes in place, so the byte count it returns is decoded bytes, not consumed bytes, we were treating it as consumed). Splits that landed inside the chunk framing went down the wrong path and returned 400.
Fixed, and added a regression test that walks every split point on the chunked body. All chunked/SSE tests green. Thanks for catching this.

PR #1062, opened 2026-07-30
[WTX] Add support for TLS stream (gRPC)

c410-f3r:
## Description

Also removes the callback that was generating logs in https://github.com/MDA2AV/HttpArena/pull/1055

PR #1297, opened 2026-08-24
[WTX] Remove gRPC

c410-f3r:
`WTX` has disabled in https://github.com/MDA2AV/HttpArena/pull/1291 because it is not passing validation tests so let's recap all my efforts to try to fix the situation:

* https://github.com/MDA2AV/HttpArena/pull/980 and https://github.com/MDA2AV/HttpArena/pull/993 failed because the project didn't have any gRPC tests. Didn't know that at the time.
* In https://github.com/MDA2AV/HttpArena/pull/1055 it became clear that CI errors were triggered by the lack of proper validation.
* https://github.com/MDA2AV/HttpArena/pull/1062 and https://github.com/MDA2AV/HttpArena/pull/1128 tried to investigate the error behind `stream-grpc-tls` without success. The integration of https://github.com/MDA2AV/HttpArena/pull/1123 didn't add much information.

In all these PRs local evaluations showed that nothing was wrong and the actual benchmark host performed its jobs in regards to unary endpoints. Worse yet, gRPC is basically (roughly speaking) HTTP/2 with protobuf still no HTTP/2 test was invalidated.

Therefore, the allegations of failed validations are dubious to say the least so let's just avoid all these inconveniences through the removal of the associated gRPC benchmarks.

EDIT: https://github.com/MDA2AV/HttpArena/pull/1291 also removed the `wtx` entry in the website across all scenarios.


MDA2AV:
Hi, I understand the feeling, gRPC has definetly gotten no love at all, mostly because I am  still  focusing on h1. 

WTX was disabled temporarily along a lot other(60+) frameworks until validation is sorted out including gRPC's which is a total mess and never worked, a lot due to the fact that there isn't much available tooling around gRPC we can just directly use. Plus the gRPC load generators are also not good and there are no good options for high throughput, h2load does not support streams and ghz results seems unreliable. 

These failed validations do not mean there is something wrong with WTX, just that until the gRPC suite is fixed, there isn't much purpose in showing results at all. WTX results will be back soon up as I work out this situation but I understand if you still want to completely remove it.

MDA2AV:
GIve me 1h, I'll try to sort this WTX situation and have it running again.

c410-f3r:
Don't worry about it @MDA2AV; the project is still in its early stages and these hurdles are expected. I understand your position, so please feel free to remove it as I don't plan to continue pursuing gRPC benchmarks.

MDA2AV:
Every WTX bench is restored and validation issues fixed to pass all checks. I would prefer continue having WTX among the results, it definitely adds value as an emerging high performance option

c410-f3r:
Thanks for the fix @MDA2AV, it was very kind of you.

Closing this PR since you want to keep the gRPC benchmarks, however, it is worth noting that they won't be part of my maintaining efforts.

PR #1312, opened 2026-08-25
update mq-bridge, add crud, fortunes

marcomq:
## Description

Updating mq-bridge to latest version.
- add crud
- add fortunes
- optimize postgres
- use mimalloc allocater

---

| Command | Description |
|---------|-------------|

```
/benchmark -f genhttp-11 --compare genhttp-11-kestrel
```

---

<details>

```bash
```

</details>


frameworks/mq-bridge/src/main.rs
```diff
@@ -1,60 +1,61 @@
...
-//! `json-tls` reuses that same `/json` handler over HTTP/1.1 + TLS on `8081`
-//! (ALPN `http/1.1` only), alongside the HTTP/2-over-TLS listener on `8443`. A
-//! cleartext HTTP/2-only (`h2c`) listener on `8082` serves the `baseline-h2c` and
-//! `json-h2c` profiles from the same handlers.
+//! Harness inputs: `DATASET_PATH`, `STATIC_DIR`, `DATABASE_URL`,
+//! `DATABASE_MAX_CONN`. A missing database is non-fatal — the DB-backed
+//! endpoints degrade rather than blocking the cleartext profiles.
+
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
```
Kaliumhexacyanoferrat: Not so sure but this *seems* to be more of a `tuned` change to me. I am not too much into Rust and `mimalloc` seems stable enough to ship to production, so I guess this is fine - just a thing to consider maybe.

marcomq: Hm. Yes - this is probably alredy a `tuned` setup. `Mimalloc` isn't completely standard - even if stable.

I mostly oriented on the trillium setup. It is also using mimalloc. Also in the not tuned version:
https://github.com/MDA2AV/HttpArena/blob/main/frameworks/trillium/src/main.rs#L2

marcomq: So - I just removed mimalloc. Will check via bench if this makes a big difference. 

Anyway - if this is an issue, trillium should probably also remove mimalloc.

marcomq:
Using `mimalloc` seems to had a significant improvement in `json-comp` for some reason. 
Keeping it disabled - let me know if it is fine to enable it.

Edit: Checking again - mimalloc also significantly increased memory consumption. So - it was probably better for throughput, but not good for the final resulted score. Will evaluate - I think I will make a lib change to improve performance without mimalloc.

frameworks/mq-bridge/src/main.rs
Kaliumhexacyanoferrat: @MDA2AV what is your take on this?

Kaliumhexacyanoferrat:
> Keeping it disabled - let me know if it is fine to enable it.

We discussed this yesterday and usage of `mimalloc` is fine for `standard` deployments.

MDA2AV (approved):

PR #1323, opened 2026-08-26
fulmine: subscribe both entries to the async profile

nigrosimone:
The async profile (#1313) is the one axis none of the other H1 profiles touch: every one of them is answered the instant it arrives, so what they separate is how fast a server can do that and not what it does with a request it cannot answer yet. GET /delay/{ms} waits the milliseconds named in its own path and echoes them back — no database, no socket, no serialization, just a timer, so the number is a property of the concurrency model rather than of somebody else's driver.

setTimeout is the framework's idiom for it and costs a timer entry: a waiting request leaves the worker free, so the 64K held connections the profile offers are 64K timers rather than 64K blocked threads.

The parameter is parsed before the timer is armed. That is not a style choice — µWS invalidates the request object the moment the handler returns, so reading it inside the callback would find it gone, and the closure is also what gives each overlapping request its own delay instead of whichever one parsed last.

millionaire needed nothing: its load is GET /baseline11?a=1&b=2, which both entries already serve, and #1319 subscribed them with the rest of the board.

Verified against the entry's own app.js on node, twice: the whole async section of validate.sh, 8 passed 0 failed each time — the random draw, the text/plain header, /delay/0 answering 0, the 10-vs-500 differential, the constant-sleep upper bound and the 32 overlapping requests with 32 distinct delays, that last one in 0.57s wall against a 499ms longest delay, so the waits overlap rather than queue. Also checked the two shapes the benchmark will hit that validation does not: 50 clients dropped mid-delay leave the workers up with a clean log, and 4000 held connections cycling /delay/15 for 5s returned 133,313 responses with no wrong body and no socket error.

No benchmark numbers here — this adds a route, it does not change one.

## Description

---

| Command | Description |
|---------|-------------|

```
/benchmark -f genhttp-11 --compare genhttp-11-kestrel
```

---

<details>

```bash
```

</details>

### DIFF
--- a/frameworks/fulmine-tuned/README.md
+++ b/frameworks/fulmine-tuned/README.md
@@ -68,6 +68,14 @@ The rest is the standard entry's, unchanged and worth repeating here:
   ordinary path, because the compiler reads the source and cannot see into one.
 - `express.compression()` is the framework's own, taking the compression module's options: it is
   mounted on the json route, which is the only one the profiles ask to compress.
+- `tls_check` is opted into in `meta.json`, and the listener it asks for on :9000 is the 8081 one
+  again, reading `/certs-tls` rather than `/certs`. What it adds is the rotation, and that is a new
+  listener rather than a new certificate on the old one: µWS reads the pair when it builds the SSL
+  context, and `addServerName()` only replaces the certificate for one SNI name — a client sending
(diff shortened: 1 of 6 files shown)


MDA2AV:
This test will be benchmarked when it is ready for all frameworks

PR #1324, opened 2026-08-26
Re-enable blackbull on 0.78.0

TOKUJI:
## Description

Re-enables `blackbull` (#1289). Bumps 0.33.1 to 0.78.0, which fixes the
fragmentation failure.

`app.py` is replaced: the previous one targets 0.33.1 and does not start on
0.78.0. `db.py` is added for `async-db`, and the Dockerfile and
`.dockerignore` now carry it into the image. The old `.dockerignore`
allow-listed three filenames, so a new module would have been silently
absent from the build.

The subscribed profile set is unchanged.

`maintainers` was empty; set to `TOKUJI`.


MDA2AV:
Pls fix the logs

TOKUJI:
Logs fixed in 0.78.1, and the entry now pins it. Two rare shutdown-time warnings are still there; the per-request ones are gone.

MDA2AV (approved):

PR #1341, opened 2026-08-26
ioxide: subscribe to async, and retune the profile to 16k/5ms

MDA2AV:
Subscribes the ioxide entry to the `async` profile, and retunes the profile itself. Both are here because #1342 was folded in on request.

## The entry

`ReactorDelay` gives the profile's wait a home that does not drag the connection off the reactor that owns it. Two mechanisms, selectable with `IOXIDE_DELAY_MODE`:

- **ring** (default) - a `timerfd` armed per request and submitted to the reactor's own ring via `SubmitRead`, so the kernel holds the deadline and the completion arrives on the reactor thread. This is how `ioxide.pg` already waits on the Postgres socket.
- **tick** - a thread-static priority queue drained from the reactor's hot path, with a sleeping fallback tick for reactors whose connections are all parked.

Measured on the bench hardware at 16,000 connections and 5ms:

| | rps | cpu | avg | p99 | p99.9 |
|---|---|---|---|---|---|
| tick | 1,842,424 | 3435% | | | |
| **ring** | **2,462,698** | 5831% | 6.48ms | 9.91ms | 16.50ms |
| tokio | 2,054,680 | 3320% | | | |

The bugs found along the way, each measured rather than guessed:

- **A process-global lock taken per request.** `Register` ran on every call. Worth **+48%** on the bench box (1,059,674 to 1,568,519), and it scales with reactor count, which is why a 32-reactor box only showed +13%.
- **A completion allocated per request.** A `TaskCompletionSource` and its `Task` was 354 MB/s of garbage, driving ~12.5ms gen0 pauses against a 15ms delay. Pooled onto the session; p99.9 halved.
- **A tick thread spinning a whole core.** `SpinOnce(-1)` never yields, so it burned 87% of a core whether it posted anything or not - the busiest thread in the process. Now below 1%.
- **The ring resuming before its deadline.** Not `EAGAIN` as expected: the fd is reused per connection, so a read could pick up an earlier request's expiration and return a valid 8 bytes. ~0.1% of waits, worst case 14.9ms into a 15ms delay. The deadline now decides, and an early completion goes back on the ring.

Things tried and reverted because they did not pay: `RecvSlots` 256 to 1024, and `RingEntries`/`ListenBacklog`/`PoolMax`/`RecvQueueEntries`, none of which moved anything beyond the run-to-run spread.

## The profile

64,000 connections at 15ms becomes **16,000 at 5ms**. Instrumenting the timer against its own deadline showed it completing 3.2ms late against 23ms of end-to-end overhead, so at 64k the profile was reporting mostly the read/write path rather than the wait. The gap between entries narrowed from 44% to 11.5% on the smaller shape.

Ceiling moves 4.27M to 3.20M. Against a best-so-far of 2.24M that is real headroom but less than before, and the docs now say so instead of claiming the ceiling does not bind.

Touches `profiles.sh`, the gcannon raw template, the CATALOG row, both profile docs and the README table. Generator runs clean, badge parity holds at 579 ranks.

**Existing `async-64000` results are orphaned by the retune**, so the column is empty until the profile is re-run for all five subscribers. The results committed here predate the retune.


MDA2AV:
/benchmark -f ioxide -t async

Measuring `e6c2e140` (drains on the reactor hot path, sleeping fallback tick, pooled completion). Not saving, this is to see whether CPU utilisation moves off 3000% of the available 6400%.

MDA2AV:
/benchmark -f ioxide -t async

Recv slots 256 -> 1024 per reactor. The timer instrumentation showed it completes only 3.2ms late on average against a 37.8ms end-to-end latency, so ~20ms is the read/write path, not the wait. This tests whether slot starvation is that 20ms.

Not saving. Watch CPU: it sat at 3000% of an available 6400%, so if this is the limiter, utilisation should rise. Memory will go up (~1GB of recv buffers vs 256MB), which is the trade being measured.

MDA2AV:
/benchmark -f ioxide -t async

Delay now defaults to the ring timer: a timerfd armed per request and submitted to the reactor's own ring via `SubmitRead`, so the kernel holds the deadline and the completion arrives on the reactor thread. No tick thread, no drain, no priority queue.

On my 32-reactor box the tick beat it (1.58M @ 1765% vs 1.50M @ 2007%), because the ring costs a `timerfd_settime` syscall plus an SQE/CQE per request while the tick batches ~13 timers per drain. But the two costs scale differently: the ring's is per request and flat, the tick's is one post per reactor per tick, so 256k/s at 64 reactors against 128k/s at 32. This reads which side wins on the hardware that matters.

Not saving. Last run was 1,568,519 @ 2802%, with 6400% available.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

Consolidated branch: profile at 16,000 conns / 5ms, ioxide on the ring timer **with the deadline guard**.

The previous 2,462,698 came from the unguarded build, where ~0.1% of waits resumed early (worst case 14.9ms into a 15ms delay). This re-reads it with those requests actually waiting, so the number should be a touch lower and honest.

tokio unchanged as the control: 2,054,680 @ 3319.6% on this profile.

Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

Profile now **32,000 connections at 10ms**. This is a control, not a retune: the ceiling is connections over delay, so this and 16k/5ms are both **3.20M**. Connection count is the only variable.

| | ceiling | ioxide | tokio |
|---|---|---|---|
| 64,000 / 15ms | 4.27M | 1,568,519 | 2,263,962 |
| 16,000 / 5ms | 3.20M | ~2,360,000 | 2,054,680 |
| 32,000 / 10ms | 3.20M | ? | ? |

The ordering reverses between the first two, but they differ in ceiling as well as connections so nothing is isolated. Here a drop against the 16k/5ms numbers is connection scaling and cannot be the arithmetic.

ioxide on the ring with the deadline guard. Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

**64,000 connections at 20ms** — third point with the ceiling pinned at 3.20M, so connection count stays the only variable.

| | ceiling | ioxide | tokio |
|---|---|---|---|
| 16,000 / 5ms | 3.20M | 2,360,000 (74%) | 2,054,680 (64%) |
| 32,000 / 10ms | 3.20M | 2,390,257 (75%) | 2,359,835 (74%) |
| 64,000 / 20ms | 3.20M | ? | ? |

The first two points already say tokio scales on this axis and ioxide does not: doubling connections moved tokio 2.05M -> 2.36M while ioxide went 2.36M -> 2.39M, flat, needing 6066% CPU against tokio 4079% to stay level.

This point tests something else. The original 64,000 run gave 1,568,519 against a 4.27M ceiling (37%), which we read as connections being the problem. Here the connections are identical and only the arithmetic differs — if this lands near 2.4M, 64,000 connections were never the difficulty.

Caveat: both are already at ~75% of 3.20M and queueing grows fast that close to a ceiling, so some convergence may be the arithmetic rather than the frameworks.

Not saving.

MDA2AV:
/benchmark-test -t async --save

Profile settled at **32,000 connections / 10ms** (ceiling 3.20M). This runs all 7 enabled subscribers and saves, repopulating the async column.

Why 32,000, from the controlled series where connections and delay doubled together so the ceiling never moved:

| | ioxide | tokio |
|---|---|---|
| 16,000 / 5ms | 2,360,000 (74%) | 2,054,680 (64%) |
| **32,000 / 10ms** | **2,390,257 (75%)** | **2,359,835 (74%)** |
| 64,000 / 20ms | 1,641,196 (51%) | 1,887,483 (59%) |

64,000 costs both entries roughly a third against an identical ceiling. Two independent servers losing together implicates something shared - the load generator holding 64,000 sockets and the kernel's cost for them - not the servers. Locally one reactor serves 1000 connections at 97% of its own ceiling on a fifth of a core, and 1000-per-reactor holds 94-97% from 1 to 16 reactors, so neither connections-per-reactor nor reactor count is what bends.

Every stored `async-<conns>` row was deleted first: 13 of them across 5 entries at 32768, 49152 and 64000, none of which the board could show. `async-db` untouched.

ioxide is on the ring timer with the deadline guard.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

ioxide now holds **one timer per reactor** instead of one per request: deadlines go in the reactor's own queue and a single timerfd is armed at the front of it, re-armed only when a new wait is due before the one it already holds. That is how Node and Bun do it, and it is the only shape that neither syscalls per request nor leaves the reactor asleep past a deadline.

The previous run is the thing to beat, and it shows why: **ioxide 2,336,243 @ 5988.1%** against **tokio 2,356,577 @ 4262.9%** — 0.9% behind on throughput while burning 40% more CPU. That CPU is the per-request `timerfd_settime` plus SQE and CQE. With waits coming due in batches of ~13, this should be about one arming per 13 requests instead of 13 armings.

Locally, 16,000 conns @ 2ms over 3 interleaved reps:

| mode | rps | cpu | p99 | p99.9 |
|---|---|---|---|---|
| queue | 1.99M | 1852% | 19.93ms | 34.23ms |
| tick | 1.98M | 1861% | 18.47ms | 31.07ms |
| ring | 1.81M | 1959% | 27.80ms | 56.03ms |

My box is saturated so the three compress together; the bench box has the headroom that made ring worth its syscall, and this should get the same wake-up behaviour without paying for it. Watch CPU more than rps.

`IOXIDE_DELAY_MODE=ring` / `=tick` still select the previous two. Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

Follow-up to the last run. The per-reactor timer was the **only** thing completing a wait, and with one wait per request the deadlines are spread continuously — so each one still got its own arm and its own read. A syscall per request, which is what the design was meant to remove. The result showed it: 2,179,118 @ 4251% with **2,150% of the box idle**.

Due waits now also complete from the reactor's own handler loop, which it passes through tens of thousands of times a second anyway. Most waits should now finish for free, leaving the timer to cover only the case where the reactor is about to sleep with work due.

Progress so far at 32,000 / 10ms:

| | rps | cpu | req/s per cpu% |
|---|---|---|---|
| ring | 2,336,243 | 5988% | 39,014 |
| queue | 2,179,118 | 4251% | 51,261 |
| queue + hot-path drain | ? | ? | ? |
| tokio | 2,366,578 | 4068% | 58,174 |

Locally this is inside the noise (2.00M @ 1877% vs 1.99M @ 1852%) because my box has no idle CPU to recover — the bench box has 2,150% of it. Watch whether rps climbs while CPU stays put.

Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

**Generator thread sweep, both servers unchanged.** `ASYNC_THREADS=128` against the 64 the profile has been running.

The profile is closed-loop: every connection holds one request for the length of its wait, so the result is connections divided by latency, and the generator is part of that latency. Both entries sit at ~13.5ms against a 10ms delay, so ~3.5ms is queueing, and both leave a third of the box idle *together* despite being unrelated implementations. Two different servers agreeing on how much CPU to leave unused points at something they share.

Locally, holding the server fixed and changing only this:

| gcannon threads | rps | server cpu |
|---|---|---|
| 8 | 1.42M | 1484% |
| 16 | 2.16M | 1987% |
| 24 | 1.92M | 1801% |

+52% with nothing on the server side touched.

Baseline to compare against, both at `-t 64`: **ioxide 2,179,118 @ 4251%**, **tokio 2,366,578 @ 4068%**.

If throughput does not move, the generator was not the constraint and the missing 3.5ms belongs elsewhere. If it does move, the profile has been reporting the harness as much as the server, and the number needs revising before a ranking is read off it.

Note this run also carries the hot-path drain commit, so ioxide has two changes in it — I will separate them if the thread count turns out to matter. Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

**Server cores halved: 32 -> 16, SMT siblings kept.** cpuset `0-31,64-95` becomes `0-15,64-79`, so 64 logical CPUs become 32 and the budget goes 6400% -> 3200%. Connections and delay unchanged at 32,000 / 10ms.

This is a decisive test of whether the idle CPU is a cap or an absence of work. The profile is closed-loop, so work is capped at connections over delay. At the measured ~19.5µs of CPU per request, running the 3.20M ceiling would take ~6240% of 6400%. ioxide sits at 2.18M — 68% of ceiling — using 4251%, which is 68% of 6240%. On those numbers nothing is blocked.

Halving the cores separates the two readings:

| outcome | meaning |
|---|---|
| drops to ~1.6M, pins near 3200% | never blocked, just out of work — the model holds |
| holds ~2.18M at ~3200% | per-request cost halved, so the wider cpuset was **losing CPU to contention** |

The second would be the interesting one: cross-reactor traffic, memory bandwidth, or SMT siblings fighting over a core. tokio is the control — if both behave the same way it is the profile, if only ioxide holds up it is ioxide-specific contention.

Current baseline at the full cpuset: **ioxide 2,179,118 @ 4251%**, **tokio 2,366,578 @ 4068%**.

Generators are already ruled out: gcannon 1.90M beats wrk 1.64M and zrk 1.39M on the same server, and tokio was flat across 64 -> 128 generator threads.

Experiment on the branch, not a proposed profile change. Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

**Same halved cpuset (`0-15,64-79`, 3200% budget), ioxide back on the per-request ring timer.**

The last run settled something: tokio pinned 3200% and ioxide did not. Identical hardware, identical ceiling, one entry able to spend everything it is given and one not. That kills the explanation I had been defending — that both were simply out of work because the profile is closed-loop. tokio was not out of work. **ioxide is blocked.**

The per-request timerfd is the only mode that has ever driven ioxide's CPU up:

| mode | rps | cpu (full cpuset) |
|---|---|---|
| ring (per-request timerfd) | 2,336,243 | **5988%** |
| queue (per-reactor timer) | 2,179,118 | 4251% |

I reverted ring earlier because 7% more throughput for 41% more CPU looked like waste. On 16 cores that judgement inverts: a mode that can occupy the machine beats one that cannot.

The mechanism fits, too. The per-request timer gives every deadline its own submission, so each wakes the reactor on its own account. The per-reactor timer holds one deadline at a time, so a reactor with work due *behind* the front of its queue stays asleep until the front fires — which is exactly the shape of a reactor that cannot fill its core.

tokio unchanged as the control. `IOXIDE_DELAY_MODE=queue` / `=tick` still select the others. Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

**ioxide now waits on `IORING_OP_TIMEOUT`.** Same halved cpuset (`0-15,64-79`, 3200% budget), 32,000 conns / 10ms, tokio unchanged as control.

The 16-core run separated the two existing modes and showed neither is good enough:

| | rps | cpu of 3200% | µs/req |
|---|---|---|---|
| tokio | 1,769,456 | **3236%** | 18.3 |
| ioxide queue | 1,335,794 | **2678%** | 20.0 |
| ioxide ring | 1,211,556 | **3192%** | 26.3 |

Per request the per-reactor timer is nearly level with tokio — it loses on **occupancy**. It holds one deadline at a time, so wake-ups come in bursts and the reactor sleeps between them with work already due: 522% of the box unused. The per-request timerfd is the mirror image — every deadline wakes the reactor, which fills the box, but a `timerfd_settime` per wait sends the CPU into syscalls instead of serving, so it delivers *less* while using *more*.

`IORING_OP_TIMEOUT` is the shape that has both: the deadline rides in the SQE, submitted with the batch the reactor was sending anyway. Per-request wake-ups, nothing spent to arrange them.

ioxide has been submitting `IORING_OP_TIMEOUT` for its own 250ms ticker all along and simply never exposed it. `Reactor.SubmitTimeout` does, in 68 lines reusing the existing slot allocation and completion routing (ioxide#212).

**Note the build:** the library package is vendored, not published — `localfeed/` holds one 140 KB build and `NuGet.config` adds it beside nuget.org, so the runner restores it from the Docker build context. Everything else resolves normally. It comes out either way once we have the answer.

What to watch: **CPU near 3200% with rps above 1.34M** means occupancy was the cap and this is the mode. CPU short of 3200% again means the idle is upstream of the timer entirely.

Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

**Back to the profile's own 32 cores (`0-31,64-95`, 6400%), ioxide on `IORING_OP_TIMEOUT`.**

The 16-core box was a diagnostic and it has answered. Every mode, with what each could actually occupy of the 3200%:

| | rps | cpu of 3200% | µs/req |
|---|---|---|---|
| tokio | 1,751,948 | 3228% | 18.4 |
| **ioxide native** | **1,422,729** | **3205%** | 22.5 |
| ioxide queue | 1,335,794 | 2678% | **20.0** |
| ioxide ring | 1,211,556 | 3192% | 26.3 |

The per-reactor timer is the cheapest of the three and still loses: cheap per request is worth nothing when 522% of the box goes unused. The per-request timerfd occupies the machine and loses harder, because what it puts the CPU to is `timerfd_settime`. `IORING_OP_TIMEOUT` occupies the machine *and* spends it on serving.

Restoring the cpuset because at 32 cores the per-reactor timer was never CPU-limited — 4251% of 6400% — so its throughput was bounded by idle rather than by hardware. That is the bound this lifts. At 22.5µs/req, 6400% is **2.84M** against tokio's 2,366,578, under a 3.20M ceiling.

That is arithmetic, not a prediction — the per-request cost may not hold with twice the reactors to keep fed. Not saving.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async --save

Final run for the record: ioxide now builds against the **released** packages, all seven at 0.7.211 plus `ioxide.timer` 0.7.211, with the vendored feed, the `NuGet.config` and the Dockerfile edits gone and `ReactorDelay.cs` deleted. The wait is one line:

```csharp
await (httpSession.Timer ??= new RingTimer(reactor)).DelayAsync(ms);
```

Same mechanism as the last run — `IORING_OP_TIMEOUT`, one timer per connection — so this is a confirmation rather than a change, and I would expect it to land where the vendored build did.

Saving this one. The committed results were taken against the vendored package, and the profile has moved to 32,000 / 10ms since the last `--save`, so the stored rows should come from a build that matches what merges.

Previous, on the vendored build: **ioxide 2,496,818 @ 5448.7%**, tokio 2,361,888 @ 4092.1%.

Verified locally against the published packages first: `/delay/10` in 10.17ms, `/delay/200` in 200.28ms, 120 sequential waits none early at 0.11ms mean overshoot, 32 concurrent distinct delays in 68.9ms against 1088ms serialised, baseline 3.85M.

PR #1342, opened 2026-08-27
async: 16,000 connections at 5ms, down from 64,000 at 15ms

MDA2AV:
Experiment, stacked on #1341 so the diff is only the profile change.

At 64,000 connections the profile turned out to be measuring mostly the read/write path rather than the wait: instrumenting the timer against its own deadline showed it completing **3.2ms late against 23ms of end-to-end overhead**, so the delay is roughly 13% of what the number reflects.

This moves the load: 16,000 connections, 5ms wait. Ceiling goes 4.27M -> 3.20M (connections over delay), against a best-so-far of 2.24M, so there is headroom but less than before and the docs now say so instead of claiming the ceiling does not bind.

Touches `profiles.sh`, the gcannon raw template, the CATALOG row, both profile docs and the README table. Generator runs clean, badge parity holds at 579 ranks.

Existing `async-64000` results are orphaned and the column is empty until a re-run. Accepted while experimenting.


MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

Profile is now 16,000 connections at 5ms (ceiling 3.20M, down from 4.27M).

Comparing the two entries under the new load. At 64,000/15ms it was ioxide 1,568,519 @ 2802% against tokio 2,263,962 @ 3970% - near-identical efficiency per unit CPU, with ioxide simply not using the box. If the gap closes here, it was connection scaling; if it holds, it is throughput.

Not saving, still experimenting.

MDA2AV:
/benchmark-multiple -f ioxide,tokio -t async

Same profile, ioxide now on the ring timer (`timerfd` per request submitted to the reactor ring) instead of the tick. tokio is unchanged and is there as the control.

Tick baseline just measured on this profile: **ioxide 1,842,424 @ 3434.9%**, tokio 2,054,680 @ 3319.6%.

At 64k/15ms the ring lost, 5% more throughput for 51% more CPU. Worth re-reading here because the profile changed underneath that result: 3x the request rate makes the ring's per-request syscall cost worse, while a quarter of the connections per reactor makes the tick's batching worse. Both point against the ring, so this is a check rather than a hope.

Not saving.

MDA2AV:
Folded into #1341 — everything is in one PR now.

The two profile commits (\`b3b554a4\`, \`ff6570b2\`) were cherry-picked across. The third commit here, the ring-timer flip, was dropped as redundant: #1341 already defaults to the ring on its own measurement, and it now also carries the deadline guard that this branch was missing.

PR #1350, opened 2026-08-27
varnish: remove thread pool overrides, use built-in defaults

gquintard:
## Summary
- `thread_pool_max=10000`/`thread_pool_min=5000` (x2 pools = up to ~20k threads) is wildly oversized for the 64-logical-CPU cpuset the benchmark container runs on for CPU-bound synthetic responses. This strips the override to get a clean varnishd-defaults baseline before sweeping custom pool sizes.
- Fixes a missing line-continuation after `vsl_mask=none` that made `-s malloc,256m` dead code, never reaching varnishd.

## Test plan
- [ ] Re-run baseline/pipelined/limited-conn benchmark profiles against this branch
- [ ] Compare RPS and `varnishstat` thread/queue counters (`MAIN.sess_queued`, `MAIN.threads_limited`) against current main


gquintard:
this confirms that tweaking the thread pools doesn't do much, and is even harmful. Closing

PR #1351, opened 2026-08-27
varnish: set thread_stats_rate=1000

gquintard:
## Summary
- Sets `thread_stats_rate=1000` (default 10) to defer per-thread worker-stat aggregation under high request rates. Experiment to isolate whether stats aggregation overhead is contributing to the baseline/pipelined RPS ceiling.
- Fixes a missing line-continuation after `vsl_mask=none` that made `-s malloc,256m` dead code, never reaching varnishd.

## Test plan
- [ ] Re-run baseline/pipelined/limited-conn benchmark profiles against this branch
- [ ] Compare RPS against current main and against the `varnish-thread-tuning` experiment (#1350)


gquintard:
well, that's surprising

PR #1353, opened 2026-08-28
capra: add Clojure framework

Ramblurr:
Adds the [capra](https://github.com/weavejester/capra) Clojure framework


Ramblurr:
I'm rescinding this PR for now, pending some upstream capra changes. Sorry for the noise, but I'll be back when it works properly.

PR #1355, opened 2026-08-28
[roda] Disable puma-io-threads

p8:
Using puma-io-threads seems to reduce performance for baseline.


p8:
/benchmark -f roda

After rebase on https://github.com/MDA2AV/HttpArena/pull/1414

p8:
/benchmark -f roda

After rebase on main which includes https://github.com/MDA2AV/HttpArena/pull/1414.

PR #1363, opened 2026-08-28
benchmark: re-measure ioxide on ioxide 0.7.211

MDA2AV:
Results-only PR, no code change.

The entry moved to **ioxide 0.7.211** across all seven packages and added **ioxide.timer**, so the async wait is now the library's `RingTimer` over `IORING_OP_TIMEOUT`:

```csharp
await (httpSession.Timer ??= new RingTimer(reactor)).DelayAsync(ms);
```

`ReactorDelay.cs` is gone with it — 552 lines at its largest, holding four competing timer mechanisms that existed to find out which belonged in the entry.

That is a library bump under **all seventeen profiles the entry subscribes to**, not only the one it was made for, so the whole set is worth re-reading rather than assuming nothing else moved. The async rows are stale for a second reason too: the profile is now 32,000 connections at 10ms, where the stored rows were taken at 64,000 at 15ms.

Verified locally against the published packages before opening this: `/delay/10` in 10.17ms, `/delay/200` in 200.28ms, 120 sequential waits none early at 0.11ms mean overshoot, 32 concurrent distinct delays in 68.9ms against 1088ms serialised, `/json/1` and a 1 MB upload unaffected, baseline 3.85M.


MDA2AV:
/benchmark -f ioxide --save

Every profile the entry subscribes to, saved.

The bump touched the whole entry, not just the async wait: all seven packages went 0.7.210 -> 0.7.211 and `ioxide.timer` was added. So this reads the full set rather than assuming only the profile the change was made for moved.

Two reasons the stored rows need replacing:

- **the library moved** under all seventeen profiles
- **`async` changed shape** — 32,000 connections at 10ms now, where the saved rows were taken at 64,000 at 15ms

For reference, `async` on the vendored build of the same code: **2,496,818 @ 5448.7%**, against tokio's 2,361,888 @ 4092.1%.

Verified locally on the published packages first: `/delay/10` in 10.17ms, `/delay/200` in 200.28ms, 120 sequential waits none early at 0.11ms mean overshoot, 32 concurrent distinct delays in 68.9ms against 1088ms serialised, `/json/1` and a 1 MB upload unaffected, baseline 3.85M.

PR #1368, opened 2026-08-29
helidon: upgrade to 4.5.4 and re-enable benchmarks

Ramblurr:
Upgrade the standard and tuned Helidon entries to 4.5.4 and re-enable them now that both pass all 87 validation checks, including exhaustive TCP fragmentation.


MDA2AV (approved):
There are tons of errors in the logs but these are not new I believe.

I'll merge this.

PR #1380, opened 2026-08-29
latency-10k: score it, and measure the field

MDA2AV:
Two things that belong together: **measure `latency-10k` across the field, and make it count.**

## Scored

The CATALOG row goes `False,True,False` → **`True,True,False`** — exactly `latency-1m`'s shape. Scored for framework entries (**flagship, emerging, experimental**) and for **engines**; not for infrastructure, matching `latency-1m`, since no proxy is measured on either profile yet.

Worth naming: the engine flag was doing nothing on its own. `scoredForType()` reads `scored` **ahead of** `engineScored`:

```js
if (t === 'infrastructure') return !!p.infraScored;
if (p.scored === false) return false;      // short-circuits engine too
if (t === 'engine') return !!p.engineScored;
return true;
```

So `False,True,False` scored the profile for *nobody*, engines included. The single `scored` flip covers both leagues.

## Measured

`/benchmark-test -t latency-10k` runs **every enabled framework subscribed to the test and saves** — **112 enabled entries**.

## Order matters, and it works out

Nothing on the board moves on the commit alone: `latency-10k` has no published rows yet, and the generator omits profiles with no rows (`if present:`). So flipping the flag is inert until the run lands.

The `--save` writes results onto this branch, which means **the flag and the data merge together** rather than leaving 112 entries with a zero in a newly-scored column.

## Docs

- the composite-score "Exception" that named only Latency-1M now covers both fixed-rate profiles and says what separates them (saturating 1M vs near-idle 10K, scored and read separately)
- the profile's own **Composite** section says scored rather than reference-only, and names which leagues
- README drops the `*`

## Verification

`node scripts/check_badge_parity.js` → *badge parity ok — 637 ranks match*. Regenerated; `latency-10k` confirmed absent from `data.js` until the run populates it.

https://claude.ai/code/session_01NsAGTadPkBtwXaYEngJomx


MDA2AV:
/benchmark-test -t latency-10k

Every enabled entry subscribed to the profile — **112 of them**.

The profile drives `GET /baseline11?a=1&b=2` at a paced 10,000 req/s on cpuset `0-31,64-95` with 1,024 connections for 20s, best-of-3 by score. Identical to `latency-1m` in every respect except the rate, which is the point: the two numbers are then directly comparable.

This run is what makes the scoring flip in this PR safe to merge — without it, 112 entries would carry a zero in a newly-scored column.

PR #1382, opened 2026-08-29
echo-100k: replace upload with a 100 KB TLS echo

MDA2AV:
[skip-maintainer-ping]

Replaces the `upload` profile with **`echo-100k`**: a 100 KB body posted over TLS and returned verbatim, so ingest and egress load at once.

## Why `upload` had to go

It measured ingest alone with bodies up to 20 MB, and had stopped discriminating:

| entry | rps @256c | ingest |
|---|---|---|
| humming-bird | 3,156 | 25.0 GB/s |
| vibe.d | 3,104 | 24.6 GB/s |
| actix | 3,062 | 24.3 GB/s |
| go-stdlib | 2,936 | 23.3 GB/s |

**A 7% spread across 99 entries** spanning D, Rust, Go, Ruby and C++ — what it looks like when a benchmark measures `memcpy` and the loopback rather than the server.

## The profile

`POST /echo` · TLS `:8081` · 100 KB · Content-Length · conns 32/256 · wrk · **reference-only**

100 KB in + 100 KB out is 200 KB per request, which leaves the bandwidth ceiling far enough away that per-request framework overhead is still visible, while being ~7 TLS records and more than one socket buffer — so partial reads, multi-record handling and partial writes all happen every request.

The endpoint is `/echo`, not `/echo-100k`, so a later profile can drive it at another size without a second route.

**Content-Length is the generator's constraint, not a preference.** wrk frames the body itself and always emits `Content-Length`; adding `Transfer-Encoding` produces a request carrying both, which [RFC 9112 §6.1](https://www.rfc-editor.org/rfc/rfc9112#section-6.1) makes an error — wrk rejects it. Verified, not assumed. Chunked moved into validation, where it is mandatory.

## Validation is byte-exact, and closes a hole the old profile had

Every `upload` check sent a body with an accurate `Content-Length` and compared the returned count — so a handler that echoed that header **without reading a byte** passed the entire suite. `echo-100k` compares bytes: 1 B, 1 KB and 100 KB random bodies, a **chunked** 100 KB body that cannot be answered from a header, and an empty body.

Every probe is pinned to `--http1.1`, and that is load-bearing: `:8081` advertises ALPN `h2`, and an echo that is broken under HTTP/1.1 can pass under h2 (see below). wrk speaks only HTTP/1.1.

The generator rotates **eight distinct 100 KB bodies** so a canned response of the right size is wrong seven times in eight.

## 106 entries implemented

Every flagship and emerging entry that can serve TLS. Seventeen are unsubscribed rather than given a listener — robyn has no TLS support at all, the WebFramework family opens a single listener so it cannot serve 8080 and 8081 together, and rage, sanic, veb and the rest simply have no second listener. Five more (araara ×2, hical, iris, typev) build from sources this repo does not contain, or are hand-written servers whose buffers are smaller than the payload.

## Five review agents, and they found real defects

Not polish — bugs that would have shipped:

- **4 Go entries were broken.** I streamed the body back using the request's `Content-Length`; `net/http` drains and closes the request body once response headers flush with unread body left (`maxPostHandlerReadBytes` = 256 KB), so the response went short under a promised length and tore the connection. The old 20 MB body was *over* that threshold, which is why the same shape worked before.
- **2 entries did not parse** — stray `});` in hono-bun and hono-node.
- **nestjs corrupted the body** — Nest's express adapter JSON-serialises a returned Buffer: 102,400 bytes out as 365,887 bytes of `{"type":"Buffer","data":[…]}`.
- **bottle returned an empty 200 on chunked** — gunicorn de-chunks but leaves `HTTP_TRANSFER_ENCODING` set, so bottle re-parsed the framing, raised 400, and a bare `except` swallowed it.
- **genhttp-kestrel did not build** — pinned to GenHTTP 10.5.1, where the `Method` enum does not exist.
- **4 entries had `/echo` on the wrong listener** (fulmine ×2, elysia, swoole, workerman) — 404 on the TLS port the profile drives.
- **2 had a duplicate Content-Type** (roda, rage) that byte comparison would never catch.

## Isolation

No regressions to other endpoints. ioxide is the only entry where the change reached shared code — its chunked decoder is also what `/baseline11` parses its integer body from — so every ioxide endpoint was re-run after: baseline GET, POST with Content-Length, **POST chunked**, pipeline, json, json+br, static, static+gzip, delay, async-db, 404, `Connection: close`, and a pipelined batch. All correct.

I also swept all changed sources for cross-endpoint edits, which caught **8 files that had silently lost their CRLF line endings** (rewriting three genhttp `Project.cs` files and both servicestack files whole). Restored.

## Smoke tested

**15 entries started and probed live, 5/5 checks each, zero failures** — all five probes over HTTP/1.1, byte-compared:

- **Go** — chi, echo, fiber, gin, go-fasthttp, go-stdlib
- **C#** — aspnet-minimal, carter, sisk, simplew, genhttp-11, fastendpoints
- **Rust** — axum, ntex, actix
- **ioxide** — plus the full endpoint regression above

The audit agent separately live-tested express, fastify, koa, node, node-h3, bun, elysia, hono-bun, hono-node, aiohttp, blackbull, django, fastapi, flask, litestar, sanic, starlette, robyn, uvicorn and fastpysgi-asgi.

## Not exercised by anyone — worth watching in this run

PHP, Ruby, JVM and the exotic tail (V, Zig, D, Lua, Luau, Perl, Julia, Swift, Elixir, Erlang, F#, C++) rest on source review plus this PR's validation. Two I would watch hardest:

- **slimeweb** — its response object exposes only `plain`/`html`/`json` plus `set_header`; there is no binary responder and none take a content type. If that build coerces to `str`, binary will not round-trip.
- **lute** — could not confirm the framework de-chunks at all, so the chunked probe is the one likely to fail there.

## Verification

`validate_profiles` 26/26, `bash -n` clean on every touched script, `node scripts/check_badge_parity.js` → *badge parity ok — 637 ranks match*. Changing `scripts/validate.sh` means this PR's validation covers **every enabled entry**, not only the ones touched.


MDA2AV:
/benchmark-multiple -f genhttp-11,genhttp-11-ioxide,ioxide,aspnet-mininal,actix -t echo-100k

MDA2AV:
/benchmark-multiple -f genhttp-11,genhttp-11-ioxide,ioxide,aspnet-minimal,actix -t echo-100k

MDA2AV:
/benchmark-multiple -f ioxide,actix,ntex,genhttp-11-ioxide -t echo-100k

MDA2AV:
/benchmark-multiple -f ioxide,actix,ntex,genhttp-11-ioxide -t echo-100k

PR #1384, opened 2026-08-30
fulmine, fulmine-tuned: nginx edge on gateway-64 and production-stack

nigrosimone:
On gateway-64 and production-stack the edge is what saturates, and the board says how much: every Caddy-fronted entry sits at 73-107k req/s with a p99 in seconds, while the nginx-fronted aspnet-minimal holds 312k and 228k with a p99 in milliseconds, on the same box and the same split class. The backend behind our Caddy edge was idling.

This moves both fulmine entries to a stock nginx:1.27-alpine edge on those two profiles, with the routing the Caddyfile had: /static/* off disk, the rest over loopback h1 keepalive, and the JWT check via auth_request before /api/*. gateway-h3 keeps Caddy, since stock nginx does not ship QUIC and on that profile every edge measures about the same. The gateway CPU split moves to the 20/12 the aspnet entry settled on.

Both configs pass nginx -t; validate-gateway is the check that exercises the compose files.


nigrosimone:
Added a commit beyond the edge swap: /baseline11 registered as get and post instead of app.all, which keeps the GET on the framework fast path (6.6% less CPU per response measured locally), and the echo route no longer copies a single-chunk body through Buffer.concat.

MDA2AV (approved):

PR #1385, opened 2026-08-30
aspnet-minimal, genhttp-11: stop buffering the echo body through a MemoryStream [skip-maintainer-ping]

MDA2AV:
Both entries collected the `/echo` body with `new MemoryStream()` before writing it back. That costs twice over: an unsized `MemoryStream` doubles as it grows (~10 allocations and a full recopy to reach 100 KB), and its final buffer lands above the **85,000-byte Large Object Heap threshold**, collected as gen2 and not compacted.

## Measured

Real profile shape — 8Gbit, 512 connections, held 50K req/s, TLS.

**aspnet-minimal** — pooled buffer sized from `Content-Length`, one read, one write:

| variant | CPU (runs) | avg |
|---|---|---|
| `MemoryStream` | 681 / 697 / 688 / 691 / 699 | **691%** |
| **`ArrayPool`** | 561 / 622 / 645 / 590 | **605%** |

**−12% CPU, and every run after beats every run before.** Requests without a `Content-Length` keep the collect-then-write path — that's what keeps the chunked case correct.

**genhttp-11** — hand the request stream straight back:

| variant | CPU | p99 |
|---|---|---|
| `MemoryStream` | 313 / 368 | 422 / 630 µs |
| **return `input`** | 347 / 348 / 350 | **227 / 268 / 276 µs** |

CPU is a wash (inside the run-to-run spread), but p99 roughly halves and steadies. Returning a `Stream` is what kept it on GenHTTP's raw response path already, so the copy was buying nothing.

## Two things this does *not* claim

- **This is not why C# trails Rust here.** The same `Echo.cs` scores **65.5 µs/req on the ioxide engine, 117.4 on GenHTTP's own, 218.9 on Kestrel** — 3.3× from the engine with the MemoryStream held constant. Pooling recovers ~12%, not the 4.7× gap to actix.
- **`[FromBody] byte[]` does not work.** Minimal APIs bind complex body types through the JSON formatter and there is no `byte[]` input formatter for `application/octet-stream`, so it answers **415**. Tested, not assumed.

## Verification

Byte-exact on both entries at 0 / 1 / 4095 / 4096 / 10240 / 102400 / 250000 bytes, plus chunked at 1 / 10 KB / 100 KB, and `/json` still serves correctly on the same keep-alive connection.

## Scope

Deliberately only these two — the ones raised. **Eleven other C# entries share the pattern** (carter 203 µs/req, genhttp-kestrel 219, fastendpoints 220, sisk 157…), as do the Go, Python and Node entries (`io.ReadAll`, `await request.body()`). Optimising a subset re-ranks the column on who got attention rather than on what each runtime costs, so the wider sweep is worth deciding as its own call.


Kaliumhexacyanoferrat:
/benchmark-multiple -f genhttp-11,genhttp-11-kestrel,genhttp-11-ioxide,aspnet-minimal -t 8gbit --save

Kaliumhexacyanoferrat:
/benchmark-multiple -f genhttp-11,genhttp-11-kestrel,genhttp-11-ioxide,aspnet-minimal -t 8gbit --save

Kaliumhexacyanoferrat (approved):

PR #1389, opened 2026-08-30
fulmine, fulmine-tuned: fulmine.js 5.18.2

nigrosimone:
fulmine.js 5.18.2 moves the recurring header names and values across the uWS boundary as cached Buffers instead of strings, which measured about 10% less CPU per response on the baseline route locally. The caret already resolves to it at build; this bumps the manifest so the run is on record against the version it measured.

### DIFF
--- a/frameworks/fulmine-tuned/package.json
+++ b/frameworks/fulmine-tuned/package.json
@@ -2,7 +2,7 @@
   "name": "httparena-fulmine",
   "private": true,
   "dependencies": {
-    "fulmine.js": "^5.18.1",
+    "fulmine.js": "^5.18.2",
     "pg": "^8.13.0",
     "pg-native": "^3.8.0",
     "pg-telaio": "^0.1.2",
--- a/frameworks/fulmine/package.json
+++ b/frameworks/fulmine/package.json
@@ -2,7 +2,7 @@
   "name": "httparena-fulmine",
   "private": true,
   "dependencies": {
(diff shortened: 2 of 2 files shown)


nigrosimone:
> 👋 Benchmark request received. A collaborator will review and approve the run.

@MDA2AV  Why do you have to approve your own requests? 🤣 set a condition so they start automatically if you are the author.

nigrosimone:
About the run saved here: it is a re-measure, not an improvement.

Every row moved between -0.5% and -2.7%, `echo-ws` and `echo-ws-pipeline` included, and those two cannot be touched by the change at all, since a WebSocket frame writes no HTTP header. So the whole sitting was about 1% slower and the rest is inside that.

What 5.18.2 does is cross the constant header names and values into uWS as cached Buffers instead of strings. It measures about 1.2us per response with the profiler, which is 3-4% of the per-request cost on this hardware, so it is below what a single run can resolve here.

Recomputed with `gen_leaderboard_data.py`, fulmine goes 4982.9 -> 4964.0 in the H/1.1 composite, same place in the field. Merge it for the manifest, not for the numbers.

nigrosimone:
Rebased on main. I dropped the results commit: #1396 trimmed the profiles after those numbers were taken, so they are for a profile set that no longer exists and CI should measure them again. What is left is the version bump on the two package.json.

MDA2AV (approved):

PR #1416, opened 2026-08-31
GenHTTP: Add more effient echo handler

Kaliumhexacyanoferrat:


Kaliumhexacyanoferrat:
No change compared to the simple "stream in, stream out" version - so the high level API is already optimized.

PR #1418, opened 2026-08-31
[rack-iodine] Update dependencies

p8:


MDA2AV:
The fact that this frameworks has 100k rps on baseline yet can achieve 1M rps on latency-1m tests tells a lot about the http parser on the POST requests

p8:
> The fact that this frameworks has 100k rps on baseline yet can achieve 1M rps on latency-1m tests tells a lot about the http parser on the POST requests

Interesting. I've created a PR that uses a different method to parse the query.

MDA2AV:
I have a feeling this is chunked parsing efficiency, has same issue before

PR #1426, opened 2026-09-02
fulmine, fulmine-tuned: publish under the package name, fulmine.js

nigrosimone:
The package is published as `fulmine.js` on npm and the repository is `fulmine.js`, but the two entries are on the board as `fulmine` and `fulmine-tuned`, so the board, the badge and the per-framework page use a name nobody searches for. This changes the `display_name` in the two `meta.json` and nothing else about the entries: same directories, same Dockerfiles, same `/benchmark -f fulmine`.

The rename moves the slug, so `site/data/results/fulmine.json` becomes `fulmine.js.json` and the `framework` field of every row goes with it. That keeps the published history attached to the entry instead of leaving it to be purged as a stale name. `scripts/gen_leaderboard_data.py` runs clean after it (no unknown-name warning, the two framework pages are written) and `scripts/check_badge_parity.js` passes, with `site/leaderboard/data.js` regenerated here.

On both entries the two POST routes now read the body with the framework's parsers, `express.text({ type: () => true })` on `/baseline11` and `express.raw({ type: () => true })` on `/echo`, instead of collecting chunks on `'data'`. The parser takes the whole body off uWS in one piece, chunked or not, and on a laptop it costs 5 to 7 us less per POST. The `type` function is there because the load generator sends no Content-Type. Checked with the validator's shapes: fragmented, lower-cased names, chunked, echo 1 B to 100 KB, chunked echo, empty body, on 8080 and 8081.

The saved run measured the parser change: baseline +11.9%, short-lived +12.6%, 8gbit 72.0 -> 68.4 us/req and p99.9 6.8 -> 4.3 ms, the other rows flat. The results commit also brought back `fulmine.json` and `fulmine-tuned.json` with the old rows, because the compare step copies main's `site/data` over the branch and never removes the files the branch renamed; they are dropped again in the last commit and `data.js` is regenerated.


nigrosimone:
Okay, the results are good for me.

MDA2AV (approved):

PR #1431, opened 2026-09-03
Re-enable Hical after fixing TCP fragmentation chunked-decode bug

Hical61:
## Background

Hical was disabled (`enabled: false`) and its results dropped in #1289 after failing the exhaustive TCP fragmentation check in #1285 — about 9% of split offsets across nine request shapes returned 400.

## Root cause

Hical uses picohttpparser for zero-copy header parsing and a custom chunked-body decoder. The old decoder tracked its position with an `encodeStart` offset and advanced it by `phr_decode_chunked`'s *output* byte count (`encodeStart += decodeBufLen`), treating output length as consumption. When a chunk frame arrived split across multiple TCP reads — with the split point inside the frame (mid chunk-size hex, between CR/LF, etc.) — the offset drifted and the decoder misreported a malformed-encoding error, answering 400.

Per the #1289 commit message, this is the "narrow slice" class of defect: an edge case at specific split points, not a parser that cannot reassemble at all.

## Fix

Fixed upstream in Hical61/Hical commit a5abb27c59e604a5fe15dfde49c634de5a13efce: the manual offset tracking was removed in favor of re-decoding unconsumed data from the front on each pass, relying on `phr_decode_chunked`'s decoder state machine to carry progress across `recv()` calls. Three fragmentation regression tests were added (`ChunkedRequestBodyFragmented`, `ChunkedRequestBodySplitEveryByte`, `ChunkedRequestWithPipelinedTail`), with `ChunkedRequestBodySplitEveryByte` mirroring this exhaustive split sweep.

## Changes

- `frameworks/hical/meta.json`: `enabled: false` → `enabled: true`
- `frameworks/hical/Dockerfile`: `HICAL_REF` → `a5abb27c59e604a5fe15dfde49c634de5a13efce`

## Verification

Locally reproduced the full post-fix request path (header parsing + `phr_decode_chunked` + pipeline tail write-back) in C and exhaustively verified all nine `validate-frag.py` request shapes at every byte offset (1,107 split points total) — zero failures, no remaining split-boundary issues.

Please re-run the benchmark and the exhaustive fragmentation validation.


Kaliumhexacyanoferrat:
Validation passed, so this worked 🙂

/benchmark -f hical --save

Hical61:
@Kaliumhexacyanoferrat  Could you please run CI again

MDA2AV (approved):

PR #1434, opened 2026-09-03
fulmine.js: one entry, with the connection headers off

nigrosimone:
`fulmine.js-tuned` existed to carry two things the standard entry could not: the file and stat caches, and `Connection: keep-alive` + `Keep-Alive` left off. The caches went in #1279, and the two headers are now what fulmine.js's own Performance tips recommend for an API behind a proxy or serving HTTP/1.1 clients (tip 10, released in 5.19.2) - the same footing the entry already stands on for `etag`. What is left between the two entries is one npm package on the two reference-only database profiles, which is not worth a second row on every board.

So this removes `frameworks/fulmine-tuned` and sets `connection headers` to false on the standard entry's three apps. A request that asks for `Connection: close` is still answered `Connection: close` and still closed; validation never looks at either header.

Checked in Docker against the validation shapes: fragmented and lower-cased requests, chunked bodies, `/echo` from 1 B to 100 KB plus chunked and empty, json-comp still gzip with the right counts and totals, on 8080 and 8081.

Two notes. The published `fulmine.js` rows were measured with the headers on, so they understate the entry until it is next benchmarked; no run is requested here. And `site/leaderboard/data.js` is regenerated, which also picks up result rows that a few other entries landed without regenerating it.


nigrosimone:
👍

MDA2AV (approved):

PR #1435, opened 2026-09-03
proxygen / proxygen-coro: shared source tree + performance work

SUJP123:
Draft. Coalesces #1097 (@lnicco) with the some refactoring work, plus optimization on top.

Both entries build from one source tree (`frameworks/proxygen/src`, selected by a `TARGET` build arg — same pattern as `sark-h3` → `sark`).

Static files are read per request and follow the disk; the precompressed `.br`/`.gz` siblings are selected off `Accept-Encoding`, which removes the on-the-fly gzip of every CSS/JS/HTML response. Response compression is scoped to `application/json`.

`validate.sh`: 82 passed, 0 failed on both entries.


SUJP123:
Wanted to benchmark these and see if results are lining up as expected before taking out of draft if thats cool.

MDA2AV (approved):

PR #1440, opened 2026-09-04
edixoi: an ioxide entry that is only the connection profiles

MDA2AV:
## Description

A second ioxide entry, `edixoi`, on the two HTTP/1.1 connection profiles only: `baseline` and `limited-conn`.

The existing `ioxide` entry serves fourteen profiles — h1, TLS, h2, h3, Postgres, static files — and the HTTP that makes that possible is ~1,900 lines across eight files. Everything in it is load-bearing for something, which is exactly what makes it a poor answer to *"what does this runtime do on a socket"*. `edixoi` is that answer, in two files:

| | |
| --- | --- |
| `Program.cs` | the transport — read, hand the bytes over, send what came back |
| `Http1.cs` | request line, query string, headers, Content-Length and chunked bodies, the response |

One endpoint, `GET|POST /baseline11?a=&b=`. Nothing above the raw recv/send API — no framework, no router — and nothing allocated per request: a chunked body's digits are accumulated as they arrive rather than reassembled, and the only thing ever copied out of the ring is a *partial* request, since a read that carries whole requests is parsed where it lies.

Configuration is two environment variables (`EDIXOI_PORT`, `EDIXOI_REACTORS`). Every ioxide knob is left at the library's own default, which is the honest setting for an entry whose point is what the runtime does unassisted — the `ioxide` entry's tuned recv buffers and write slab came from profiles this one does not run.

It sits beside the existing entry rather than replacing it; `frameworks/ioxide` is untouched.

## Validation

`./scripts/validate.sh edixoi` → **22 passed, 0 failed**, including the exhaustive fragmentation sweep:

```
[frag] 9 request shapes, 1047 split offsets, 200ms pause per batch of 384
  PASS [frag: GET, minimal]                      all  73 offsets -> 200 '55'
  PASS [frag: GET, several headers]              all 148 offsets -> 200 '55'
  PASS [frag: GET, lower-cased field names]      all 101 offsets -> 200 '55'
  PASS [frag: GET, randomized query]             all  75 offsets -> 200 '1224'
  PASS [frag: POST, Content-Length]              all 121 offsets -> 200 '75'
  PASS [frag: POST, lower-cased content-length]  all 121 offsets -> 200 '75'
  PASS [frag: POST, randomized body]             all 123 offsets -> 200 '1311'
  PASS [frag: POST, chunked]                     all 140 offsets -> 200 '75'
  PASS [frag: POST, chunked in two chunks]       all 145 offsets -> 200 '75'
=== Frag Results: 9 passed, 0 failed ===
```

Keep-alive, pipelining (three requests in one write, three correct answers) and `Connection: close` were checked separately over a socket.

## Two things the sweep caught

Recorded because neither is visible from a request that arrives whole, and both looked like a working server until the split landed in the wrong place.

**`Connection: close` needs an explicit `shutdown(fd, SHUT_WR)`.** Returning from the handler is not enough: ioxide closes the descriptor when its own reference goes and the multishot recv still holds one, so the peer sat waiting for a FIN that had not been sent, and every close-delimited request timed out. `SHUT_WR` rather than `close(2)` — the descriptor's lifetime is the runtime's, and closing it here would free a number the reactor still holds, which the next accept could hand to somebody else.

**The close must not be applied until the whole request is in hand.** Reading it off the headers meant a POST whose body had not arrived yet got its socket half-closed mid-request. 5 of the 9 shapes failed, at exactly the offsets that split after the header and before the body — the GET shapes passed throughout, because for them head-complete *is* request-complete.

## Local numbers

Not arena results — a 32-core dev box with wrk on the same machine, GET only, so they say "not pathologically slow" and nothing more: **~1.37M rps** on 8 reactors, 256 connections, p50 ~101µs.

Happy to have `/benchmark -f edixoi --compare ioxide` run for the real ones.


MDA2AV:
Now subscribed to `latency-1m`, `latency-10k`, `async`, `json-tls` and `8gbit` as well.

The two fixed-rate profiles were free — both drive `GET /baseline11`, so subscribing was the whole change. The other three needed endpoints: `GET /delay/{ms}` on ioxide.timer (the wait is an `IORING_OP_TIMEOUT`, so a held connection costs a kernel deadline rather than a thread), `GET /json/{count}?m=N` serialized per request from the dataset, and `POST /echo` returning the body byte for byte from either Content-Length or chunked framing. The last two brought TLS on :8081 — one `ExtraPorts` entry and a `TlsService` per reactor, with the mounted certificate used as-is.

**50 passed, 0 failed**, including the TLS quality and posture probes, 32 concurrent delays each with its own deadline, json-tls checked field by field against `data/dataset.json`, and every echo compared byte for byte at 1B–102400B.

Note the saved results in `db41023` (baseline 4,112,572 · limited-conn 2,699,677) predate this commit — they measured the baseline-only build. Worth re-running now that the entry serves seven profiles.

PR #1441, opened 2026-09-04
ioxide-pipes: the same two profiles over the pipe API, parsed by Glyph11

MDA2AV:
## Description

A companion to [#1440](https://github.com/MDA2AV/HttpArena/pull/1440) (`edixoi`) rather than a replacement for it: same runtime, same two profiles (`baseline`, `limited-conn`), same endpoint — and neither of the two things `edixoi` hand-writes.

**The pipe carries the design.** Its reader owns the carry, so a request split across several recvs is handed over as one `ReadOnlySequence<byte>` with no copy and no buffer of the entry's own. `AdvanceTo(consumed, examined)` is the *whole* of the partial-request handling, where `edixoi` needs a `Carry` class to hold the tail between reads. The writer stages into the connection's slab, so a response still leaves in one send.

**Glyph11 takes that sequence as it is.** It owns the two places a hand-written parser gets fiddly — the query string, handed back already split as `BinaryRequest.QueryParameters`, and chunked framing, streamed a chunk at a time through `ChunkedBodyStream`'s multi-segment overload with no linearization. What's left here is a transport loop and a routing decision.

| | |
| --- | --- |
| `Program.cs` | the transport — `ReadAsync`, hand the sequence over, `AdvanceTo`, flush |
| `Http1.cs` | what to do with what Glyph11 parsed |

`BinaryRequest` is cleared and refilled per request rather than allocated. The one place this entry *does* allocate is worth naming rather than glossing: a head that straddles a read boundary is linearized with `ToArray()` inside `TryExtractFullHeader` before parsing.

Configuration is two environment variables (`IOXIDE_PIPES_PORT`, `IOXIDE_PIPES_REACTORS`); every ioxide knob is left at the library's own default, same as `edixoi`.

## Validation

`./scripts/validate.sh ioxide-pipes` → **22 passed, 0 failed**, including the exhaustive fragmentation sweep:

```
[frag] 9 request shapes, 1047 split offsets, 200ms pause per batch of 384
  PASS [frag: GET, minimal]                      all  73 offsets -> 200 '55'
  PASS [frag: GET, several headers]              all 148 offsets -> 200 '55'
  PASS [frag: GET, lower-cased field names]      all 101 offsets -> 200 '55'
  PASS [frag: GET, randomized query]             all  75 offsets -> 200 '1424'
  PASS [frag: POST, Content-Length]              all 121 offsets -> 200 '75'
  PASS [frag: POST, lower-cased content-length]  all 121 offsets -> 200 '75'
  PASS [frag: POST, randomized body]             all 123 offsets -> 200 '1496'
  PASS [frag: POST, chunked]                     all 140 offsets -> 200 '75'
  PASS [frag: POST, chunked in two chunks]       all 145 offsets -> 200 '75'
=== Frag Results: 9 passed, 0 failed ===
```

Keep-alive, pipelining and `Connection: close` were checked separately over a socket.

## Local numbers — the pipe adapter is not visible at this load

8 reactors, `wrk -t4 -c256`, one listener asserted each time. GET only, dev box, generator on the same machine, so this is a smoke test and not a result:

| | run 1 | run 2 |
|---|---|---|
| `edixoi` (raw API) | 1.393M | 1.381M |
| `ioxide-pipes` (pipe API + Glyph11) | 1.392M | 1.339M |
| `edixoi` re-measured as a control | 1.381M | 1.370M |

The spread *within* one entry is as wide as the spread *between* the two, so at this load on this rig they are the same number. The bench rig's mixed GET/POST/chunked rotation at 4,096 connections is the comparison that will actually say something — `/benchmark -f ioxide-pipes --compare edixoi` is the run worth having, once #1440 is in.

## One bug worth recording

A POST answered `5520` for `?a=13&b=42` with body `20`. The body's digits were being folded into the *running* total instead of accumulated on their own and added, so every body byte shifted the query sum a decimal place: `55 → 552 → 5520`. The chunked path had it right, which is what made the two disagree and gave it away.


MDA2AV:
Heads up on the saved numbers: the `/benchmark --save` run in `a4a1d1a` (baseline 3,653,796 rps · limited-conn 2,691,138 rps) was measured **before** `runtime-async=on` went on the csproj in `7dc1b73`.

The `ioxide` entry has carried that flag since it landed and this one was written without it, so the two were not being compiled the same way. Every request here is served from an async method awaiting an `IValueTaskSource` that completes inline on the reactor — exactly the shape runtime async consumes as-is — so the flag is not cosmetic on this entry.

Worth re-running before these numbers go on the board. Revalidated with it on: 22 passed, 0 failed, fragmentation sweep included.

MDA2AV:
Now subscribed to `latency-1m`, `latency-10k`, `async`, `json-tls` and `8gbit` as well — same endpoints as [#1440](https://github.com/MDA2AV/HttpArena/pull/1440), and the TLS door is where the pipe seam earns its place: `TlsConnectionDualPipe` is an `IDuplexPipe` like any other, so the serve loop is written once and never learns which one it got, where the raw entry threads a session through every read and every write.

The echo plays to the sequence API too — the body goes back out of the reader's own segments rather than through a buffer of ours, chunked or not, because Glyph11's multi-segment `TryReadChunk` decodes in place.

**50 passed, 0 failed.**

Note the saved results in `c9f3d4e` (baseline 3,901,974 · limited-conn 2,653,002) predate this commit — they measured the baseline-only build. Worth re-running now that the entry serves seven profiles.

MDA2AV:
Found and fixed it — thanks for flagging.

`Http1` was constructed **per connection**, and the endpoint work gave it a `Growable` that took 16 KB up front. `limited-conn` closes every connection after ten requests, so that is a quarter of a million 16 KB allocations a second — which is exactly the shape of the result: limited-conn −23.6% and memory +14%, baseline flat.

It never needed to be per connection. `Serve()` does not await — the delay endpoint hands the milliseconds back to the caller rather than waiting inside the parser — and a reactor runs a single thread, so no connection can be inside `Serve` while another is. It is a reactor service now, built once per thread, and a connection allocates nothing to be parsed. The scratch buffer starts empty too.

Benchmarked locally before pushing, `wrk -t4 -c256` with `Connection: close` — the same per-connection cost limited-conn pays and nothing else:

| build | run 1 | run 2 |
|---|---|---|
| per connection (the regression) | 224,204 | 223,950 |
| **per reactor (this)** | **245,856** | **247,600** |
| per reactor, re-measured as control | 247,732 | 248,656 |
| `7dc1b739e`, the build that scored 2,653,002 | 245,599 | 247,135 |

So it is back at parity with the build the good number came from, not merely better than the regression. Keep-alive is untouched either way — 1,450,927 / 1,436,004 before, 1,431,885 / 1,452,853 after — which is what the flat baseline already said.

**50 passed, 0 failed**; the whole suite was re-run rather than the endpoints spot-checked, since sharing the parser across a reactor's connections is a real change.

[#1440](https://github.com/MDA2AV/HttpArena/pull/1440) had the same fault and worse — two 16 KB buffers plus an 8 KB carry, 40 KB per accept, **+28%** once fixed — caught before its expanded build was ever benchmarked. Both are ready for a re-run.

PR #1442, opened 2026-09-04
fulmine: drop the error middleware, it was taking the fast paths off every route

nigrosimone:
The entry registered `app.use(answerError)`, a four argument handler, only to answer an empty body
on the errors `express.json()` and `express.static` raise. fulmine grants a route its fast paths
by reading the source of every callback in its chain, and a four argument handler is code it cannot
follow: one anywhere in the app takes the grant off every route. So the entry has been paying the
request header copy on all of its GET routes, and the query fetch on the ones that never read it,
including `/baseline11`, whose own comment says it does not.

The fix is to delete it. `npx fulmine profile app.js` inside the container, before and after:

```
before:  0 routes copy no request headers
after:   9 copy no request headers, 4 read no query
```

Nothing is given up. The framework's final handler answers with the status the error carries, so
`/static/nonexistent.txt` is still 404, a broken JSON body still 400 and a body over the limit still
413; only the body changes, from empty to an error page, and no profile sends a request that fails.
Checked in Docker against the current entry side by side: 19 requests, no status differs and every
non-error response is byte identical.

Not asking for a run: the published rows were measured with the copy on, so they understate the
entry until it is remeasured.


nigrosimone:
/benchmark --save

The published rows were measured with the header copy on every GET, so they understate this entry until it is remeasured. The GET profiles are where the change should show: baseline, latency-1m, latency-10k, limited-conn, json-comp and json-tls.

MDA2AV (approved):

PR #1445, opened 2026-09-04
[bun] serve the h2, h2c and h3 profiles from the same entry

ReneWerner87:
`Bun.serve` grew two experimental flags: `http2: true` (1.4.1, oven-sh/bun#40137)
and `http3: true` (1.4.0, oven-sh/bun#29768). Both are a property of the listener
rather than a second server, so this entry picks up six more profiles off the
routes it already has - no second implementation, no proxy in front.

| Port | Transport | Protocol | Profiles |
|------|-----------|----------|----------|
| 8080 | TCP | HTTP/1.1 | the twelve h1 profiles |
| 8081 | TLS | HTTP/1.1, no ALPN negotiated | `json-tls`, `static-tls`, `8gbit` |
| 8082 | TCP | h2c on the h2 preface | `baseline-h2c`, `json-h2c` |
| 8443/tcp | TLS | h2 via ALPN | `baseline-h2`, `static-h2` |
| 8443/udp | QUIC | HTTP/3, Bun's own lsquic | `baseline-h3`, `static-h3` |

:8081 deliberately carries no `http2` flag: its three profiles are measured over
HTTP/1.1, and the flag would hand h2 to any client whose ALPN offers it. :8082
dual-serves HTTP/1.1 next to h2c, which baseline-h2c permits outright, and neither
of its drivers can reach that path - both open with the preface.

Every listener past :8080 is wrapped, so a bind failure costs only its own profiles
instead of taking the twelve h1 ones down with it.

The base image is pinned via `ARG BUN_VERSION=1.4.1`. 1.4.0 ignores `http2` rather
than rejecting it, so :8082 and :8443 would quietly answer HTTP/1.1.

`maintainers` is set to Bun's own, per oven-sh/bun's CODEOWNERS: `Jarred-Sumner`
owns the project, `alii` owns `packages/bun-types` and `*.d.ts`, where the two
serve flags this entry depends on are typed.

### Verification

Run in the container on 1.4.1 with 10 workers, against the harness's own probes:

- 14/14 on h1, h2c, json-h2c, h2, static-h2 content types and 404, and the :8081
  TLS posture (no ALPN negotiated, TLS 1.3)
- `baseline-h3` and `static-h3` at 4/4 2xx with `Application protocol: h3`, through
  the repo's own ngtcp2-built `h2load`
- all four workers bind tcp/8443 and udp/8443 through `reusePort`


MDA2AV (approved):

ReneWerner87:
@Jarred-Sumner good work

image
image
image
image
image

PR #1448, opened 2026-09-05
Add the neton entry for Kotlin Native

zoujiaqing:
## Description
Adds Neton, a Kotlin Multiplatform framework compiled to a native executable.

Routing, middleware, request parsing and the response pipeline are the framework's.
The hyper4k engine — Tokio + Hyper 1.x, linked in as a Rust static library — does
protocol and transport only. The runtime image carries no JDK: Kotlin/Native
compiles to a native binary, so the JDK exists in the build stage alone.

Two listeners share one route table: `:8080` for the HTTP/1.1 profiles and `:8082`
for the h2c ones. hyper4k answers HTTP/1.1 and HTTP/2 cleartext on the same socket
by prior knowledge, so the second listener is a second adapter over the same frozen
application context, not a second application.

The framework is consumed from Maven Central as a single coordinate,
`com.netonstream:neton:1.0.0-beta5`, which brings core, logging, HTTP, routing and
the engine. The entry therefore builds exactly the way an application would — no
source checkout and no composite build, so what is measured is what a user gets.

**Subscribed:** baseline, limited-conn, pipelined, async, latency-1m, latency-10k,
baseline-h2c, json-h2c.

**Not subscribed:** the TLS profiles, because the engine terminates no TLS today,
and json-comp, which needs gzip/br response compression.

**One configuration note.** `config/application.conf` sets `timeout = 0`, which
disables the per-request timeout wrapper. That wrapper allocates two coroutine
objects and their state transitions on every request, and each arena profile
already carries its own deadline, so a second one on the server buys nothing here.
Shutdown draining is unaffected — the adapter keeps its own 5s grace. Say the word
and I will restore the framework default of 30s.

**Disclosure:** I have not been able to run the Docker build locally — no Linux
Docker host available on my side — so CI is its first real execution. Two things
could plausibly fail there: downloading the Kotlin/Native toolchain inside the
container, and the Kotlin 2.4.20-RC3 compiler consuming klibs published with
2.4.0. I will push fixes to this branch if either bites.

---

| Command | Description |
|---------|-------------|

```
/benchmark -f genhttp --compare genhttp-kestrel
```

---

<details>

```bash
```

</details>


MDA2AV (approved):

frameworks/neton/src/nativeMain/kotlin/Main.kt
```diff
@@ -0,0 +1,197 @@
...
+    response.json(items.render(request.pathParam("count"), request.queryParam("m")))
+}
+
+/**
+ * The dataset behind /json/{count}?m=M: the first `count` items of
+ * /data/dataset.json, field-for-field unchanged, each with
+ * total = price * quantity * m added.
+ *
+ * Parsed once at startup and kept as the JSON text that *precedes* each total,
+ * so serving a request appends one number per item and re-serializes nothing.
```
MDA2AV: rules state that you need to serialize every request.

PR #1452, opened 2026-09-05
Add a Watson Webserver entry

MDA2AV:
[Watson Webserver](https://github.com/jchristn/WatsonWebserver) — a small async C# web server for REST endpoints.

Written the way a caller writes it: routes are declared through Watson's own routing — static routes for the fixed paths, parameter routes for the two that carry a value in the path — rather than a hand-rolled dispatch, because that is what the library is for and what its numbers should reflect.

## Scope

HTTP/1.1 cleartext on `:8080`, plus the async delay. No TLS listener, so the profiles measured over one (`json-tls`, `8gbit`, the static ones) are not claimed.

| Profiles |
|---|
| baseline, pipelined, limited-conn, async, latency-1m, latency-10k, json-comp |

## Notes

- `/json` is serialized from the parsed model on **every** request and compressed only when the client asked for it, at brotli quality 1. Nothing is answered from a precomputed body — the rule the board's anti-cheat check exists to enforce.
- `/delay/{ms}` uses `Task.Delay`, which registers a timer and yields rather than holding a thread, so waits in flight are bounded by memory.
- The project is named `watsonarena` rather than `watson`. A project named for the package it depends on makes NuGet resolve a dependency cycle on itself: `NU1108: Cycle detected. watson -> Watson (>= 7.1.1)`. Worth knowing, since the message doesn't say what to do about it.

## Validation

`./scripts/validate.sh watson` — **37 PASS, 0 FAIL**.


MDA2AV:
Added a TLS listener on `:8081` and the `/echo` route, so the entry now also claims **json-tls** and **8gbit** — both are measured over TLS, which is the only reason they were missing.

Watson terminates TLS itself through its own `SslSettings`, and both listeners answer the same routes; only the transport underneath differs.

Two things worth noting for anyone doing the same:
- `StartAsync` does not return while the listener is running, so awaiting it means a second listener never comes up — and nothing is logged, you just get one socket. `Start()` is the non-blocking one.
- `CreateFromPemFile` leaves the private key ephemeral, which the TLS stack won't accept; it needs round-tripping through PKCS#12.

Validation is now **57 PASS, 0 FAIL** (was 37).

Issue #91, opened 2026-03-24
[Audit] spring-jvm: /compression uses default gzip level (~6) instead of required level 1


jerrythetruckdriver:
## Framework
spring-jvm (Spring Boot + Tomcat)

## File
`src/main/resources/application.properties`

## What it does
The compression endpoint at `/compression` relies on Spring Boot's built-in Tomcat compression:

```properties
server.compression.enabled=true
server.compression.mime-types=application/json
server.compression.min-response-size=1024
```

The handler in `BenchmarkController.java` (line ~100) simply returns the pre-serialized `largeJsonResponse` byte array and lets the framework handle compression:

```java
@GetMapping(value = "/compression", produces = MediaType.APPLICATION_JSON_VALUE)
public byte[] compression() {
    return largeJsonResponse;
}
```

Tomcat's compression uses Java's `GZIPOutputStream` which defaults to `Deflater.DEFAULT_COMPRESSION` (internally mapped to zlib level 6). Spring Boot's `server.compression.*` properties do not expose a compression level setting.

## What the spec requires
Gzip level 1 (fastest). The benchmark profile specifies level 1 to keep the comparison fair across frameworks — higher compression levels trade CPU for better ratios, which skews throughput numbers.

## Impact
Using level 6 vs level 1 means significantly more CPU per response spent on compression. This gives spring-jvm an unfair disadvantage in throughput (slower per-request) but an unfair advantage in bandwidth (smaller responses). Either way, it's not measuring what the benchmark intends.

## Suggested fix
Since `server.compression.*` doesn't support level config, replace it with a manual gzip filter or handle compression directly in the endpoint:

**Option A — manual gzip in handler:**
```java
import java.util.zip.GZIPOutputStream;
import java.util.zip.Deflater;

@GetMapping(value = "/compression", produces = MediaType.APPLICATION_JSON_VALUE)
public void compression(HttpServletRequest req, HttpServletResponse resp) throws IOException {
    String ae = req.getHeader("Accept-Encoding");
    if (ae != null && ae.contains("gzip")) {
        resp.setHeader("Content-Encoding", "gzip");
        resp.setContentType("application/json");
        GZIPOutputStream gos = new GZIPOutputStream(resp.getOutputStream()) {{
            def.setLevel(Deflater.BEST_SPEED); // level 1
        }};
        gos.write(largeJsonResponse);
        gos.finish();
    } else {
        resp.setContentType("application/json");
[...]

PR #38, opened 2026-03-16
Add Axum: ergonomic Rust web framework built on Tokio/Hyper (~20k⭐)

BennyFranciscus:
## Axum

[Axum](https://github.com/tokio-rs/axum) is the most popular modern Rust web framework, built on top of [Hyper](https://github.com/hyperium/hyper), [Tokio](https://tokio.rs/), and [Tower](https://github.com/tower-rs/tower). It's become the go-to choice for Rust web services.

### Why add it?

HttpArena already has **actix**, **hyper**, and **salvo** for Rust — but Axum is the framework most Rust developers are choosing today. It's the missing piece for a complete Rust comparison:

- **hyper** — the raw HTTP library (already in HttpArena)
- **actix** — the actor-based framework (already in HttpArena)
- **axum** — the ergonomic, extractor-based framework (**this PR**)
- **salvo** — the feature-rich framework (already in HttpArena)

This gives people a really interesting comparison: raw hyper vs actix vs axum vs salvo, all in Rust, with very different design philosophies.

### Implementation details

- Axum 0.8 on Tokio multi-threaded runtime
- Rustls for TLS/HTTP2 via `axum-server`
- `tower-http` CompressionLayer for gzip
- Pre-opened SQLite connection pool (one per CPU core) with mmap
- Static files and datasets loaded into memory at startup
- Compiled with `-O3`, thin LTO, single codegen unit, panic=abort

### Endpoints

All standard HttpArena endpoints: `/pipeline`, `/baseline11`, `/baseline2`, `/json`, `/compression`, `/db`, `/upload`, `/static/{filename}`

PR #1457, opened 2026-09-06
fixed-rate scoring: mean latency replaces p99.9, plus a latency-500k-8cpu profile on four cores [skip-maintainer-ping]

MDA2AV:
## Why

On latency-1m the eight cheapest entries land within 2 µs of each other (18.7–20.7 µs/req) with 84–137 µs mean latency: at 1M req/s over 64 threads every thin server pays one wake-up per request, and they all pay the same one. The only way under that floor is to let requests queue and handle several per wake-up, which is what genhttp does (482 µs mean, 3.4 ms p99, and a cost that falls 1.9x between 10K and 1M req/s while ioxide's falls 1.1x). The generator is not the cause: cross-checked against wrk2, which agrees within 10% on ioxide and would flatter genhttp by 20–30%, since its millisecond timer wheel clumps sends. The CPU number is right; what it is allowed to mean is the fix.

## Scoring (latency-1m, latency-10k, 8gbit, and the new profile)

```
quality   = 0.50 × cpuScore + 0.25 × p99Score + 0.25 × meanScore
meanScore = 1 − log10(mean / bestMean)         one decade: 10× the best is 0
p99Score  = 1 − log10(p99 / bestP99) / 3       unchanged
```

The mean is the queueing signal (in-flight requests over the rate). Among entries that answer without queueing it spans about 2x, so a mean ten times the best is worth nothing. p99.9 stays on every row and in the cell popup but is no longer scored: one stall in twenty seconds moves it a hundredfold (tokio: 166 µs p99 next to a 25 ms p99.9), and at the weight it carried that noise was deciding ranks.

Effect on the published boards, current → new:

| latency-1m | | latency-10k | |
|---|---|---|---|
| zeemo 97 (#1) → 98 (#1) | | tokio 94 (#2) → 98 (#1) | |
| ioxide 96 (#2) → 96 (#3) | | ioxide 96 (#1) → 96 (#2) | |
| tokio 90 (#6) → 94 (#7) | | genhttp 69 (#31) → 72 (#33) | |
| genhttp 71 (#17) → 62 (#28) | | aspnet-minimal 37 (#91) → 50 (#72) | |
| axum 61 (#26) → 49 (#38) | | | |

latency-10k reshuffles the most (58 entries move ≥3 places) because p99.9 there was mostly single stalls. 8gbit barely moves.

## latency-500k-8cpu

`GET /baseline11` at a pinned 500K req/s, 1,024 connections, 20 s, with the server on cpuset `0-3,64-67`: four cores and their SMT siblings. Near saturation for most entries, so the marginal cost per request and the queue it builds are what get measured rather than the wake-up floor. Unscored in every league while the first field is measured. Subscribed on ioxide, genhttp-ioxide, genhttp, bun and actix.

It started as two cores (`0-1,64-65`). The first run on the bench host showed that was a capacity test, not a cost test: ioxide held 54% of the rate (271K req/s at 388% CPU, 4.5 s mean), genhttp 34% (172K, 6.6 s mean), and actix, bun and genhttp-ioxide produced no row at all. Those rows were dropped with the rename. Scaling from that run, eight logical CPUs give ioxide about 540K of headroom on that host, so the rate is now reachable for the fastest entries and everyone else's shortfall is a measurement rather than a wall.

Local preview on an i9-14900K (CPUs 0-7 = four P-cores with SMT), same formula:

| entry | held | µs/req | CPU used | mean | p99 | score |
|---|---|---|---|---|---|---|
| actix | 1.00 | 6.1 | 38% | 43 µs | 38 µs | 90.0 |
| genhttp-ioxide | 1.00 | 5.6 | 35% | 49 µs | 555 µs | 82.5 |
| ioxide | 1.00 | 4.9 | 31% | 208 µs | 84 µs | 80.0 |
| bun | 1.00 | 10.1 | 63% | 46 µs | 252 µs | 66.7 |
| genhttp | 1.00 | 7.9 | 49% | 432 µs | 2.1 ms | 41.6 |

ioxide's mean there is a stall, not queueing: p50 20 µs, p99 84 µs, p99.9 42 ms on a desktop that was not idle. It is a fair reminder that the mean is more outlier-sensitive than p99; best-of-three and a dedicated host are what keep that in check.

## genhttp-ioxide

The `ioxide-0.4.161` branch it cloned is merged upstream and gone, so the image could not be built anywhere. It now builds from the `11.0.0-preview.24` packages like the genhttp entry (Engine.Ioxide pins ioxide 0.13.218, matching the existing `ioxide.timer` pin). Builds clean; `validate.sh` passes 107/107 locally including h2, h2c, h3, WebSocket, async-db, 8gbit and the TLS verdict.

## Verification

- `python3 scripts/gen_leaderboard_data.py` and `node scripts/check_badge_parity.js` pass (671 ranks match).
- `bash -n` on the four shell files, `py_compile` on the two Python files.
- The `=== Fixed-rate:` line in benchmark.sh still prints p99.9 because `benchmark-pr.yml` parses that exact shape; switching both to the mean is a small follow-up.

## To run

```
/benchmark-multiple -f ioxide,genhttp-ioxide,genhttp,bun,actix -t latency-500k-8cpu --save
```

[skip-maintainer-ping]

https://claude.ai/code/session_01UnZzwv29UoYNVCUq24YeJF


MDA2AV:
/benchmark-multiple -f ioxide,genhttp-ioxide,genhttp,bun,actix -t latency-500k-4cpu --save

MDA2AV:
/benchmark-multiple -f ioxide,genhttp-ioxide,genhttp,bun,actix -t latency-500k-8cpu --save

PR #1473, opened 2026-09-09
libioxd: the libioma entry under the library's new name, on every engine-scored H1 profile plus static-tls

MDA2AV:
## Summary

The engine behind the `libioma` entry was renamed **libioxd** ([github.com/MDA2AV/libioxd](https://github.com/MDA2AV/libioxd), [manual](https://mda2av.github.io/libioxd/)) and its API changed, so the entry follows: directory, `meta.json`, the Dockerfile's pinned commit (`LIBIOXD_VERSION`) and the handlers.

The entry now subscribes to every H1 profile an engine is scored on - `baseline`, `limited-conn`, `async`, `latency-1m`, `latency-10k`, `json-tls`, `8gbit` - plus `static-tls`, and keeps the `latency-500k-8cpu` reference it already ran.

| Endpoint | Profile docs | How |
|---|---|---|
| `GET/POST /baseline11` | [baseline](https://www.http-arena.com/docs/test-profiles/h1/baseline/implementation/) | query slices summed, the body read on demand, digits written into the reply slab |
| `GET /delay/{ms}` | [async](https://www.http-arena.com/docs/test-profiles/h1/async/implementation/) | `ioxd_delay`: the kernel's timeout on the ring, one entry per request; the coroutine parks |
| `GET /json/{count}?m=` | [json-tls](https://www.http-arena.com/docs/test-profiles/h1/json-tls/implementation/) | dataset parsed once with cJSON, serialized per request by the library's JSON writer, `total = price × quantity × m` |
| `POST /echo` | [8gbit](https://www.http-arena.com/docs/test-profiles/h1/8gbit/implementation/) | the body read from its framing (Content-Length or chunked) straight into the reply slab and sent as it is read |
| `GET /static/{file}` | [static-tls](https://www.http-arena.com/docs/test-profiles/h1/static-tls/implementation/) | opened, sized and read from disk on every request, nothing cached; the `.br`/`.gz` twin when `Accept-Encoding` takes it |

TLS on `:8081` is TLS 1.3 only: OpenSSL runs the handshake, the kernel frames the records (kTLS transmit and receive), so the bench host needs the `tls` module as it does for the ioxide entry. libioxd loads a `<dir>/<host>/cert.pem` store, so the entry links a `default` host to the mounted `/certs/server.crt` and `server.key`.

The entry's rows in `site/data` (results, `frameworks.json`) carry the new name so the board does not keep a `libioma` ghost; the numbers are the old ones until the next benchmark run.

## Validation

`./scripts/validate.sh libioxd` on the final build: **63 passed, 0 failed** - baseline (including the exhaustive fragmentation sweep, 9 shapes × 1047 offsets), async (32 overlapping delays), json-tls (items matched against the dataset), 8gbit (byte-exact at 1 B to 100 KB, Content-Length and chunked, empty body), static-tls (20 sizes, 15 compressed twins, both follow-the-disk probes), and the TLS posture and quality probes on `:8081`.

https://claude.ai/code/session_013wYnJvEFUjKEpGLkyTLt9P


MDA2AV:
CI failed only the two 100 KB echoes (0 bytes back), and it was a library bug rather than an entry one. The streamed echo sends each piece while the rest of the body is still arriving; on the hosted runner the body came in as small completions, enough to fill libioxd's 64-slot per-connection receive queue while the handler was parked on its first send, and the library's policy for a full queue was to end the input and reset the peer. Locally the same body arrives in 2 KB completions and never got near the mark, which is why validation passed here twice. Reproduced with a 4-deep queue: a 100 KB echo died after 16 KB.

Fixed in libioxd ([85b61bd](https://github.com/MDA2AV/libioxd/commit/85b61bd6ccd4e3c6638a882d3d663012af23eddf)): at the mark the multishot recv is paused, the socket's own window holds the peer, and the reader re-arms it as it drains; the burst the kernel had already posted is kept in a grown ring, so nothing delivered is dropped. The library's suites gained a 1 MiB streamed echo (also through a 4-deep queue under buffer starvation). The entry is unchanged; only the pinned commit moves. `scripts/validate.sh libioxd` on the new pin: 63 passed, 0 failed.

MDA2AV:
potential ENOBUF buffer starvation in some tests, investigate.

PR #1483, opened 2026-09-18
[feat] hical: add echo-ws-pipeline and echo-ws-limited WebSocket tests

Hical61:
## Changes

Add two WebSocket profiles for Hical: `echo-ws-pipeline` (16-message back-to-back echo throughput) and `echo-ws-limited` (short-lived connection churn).

Changes to `frameworks/hical/meta.json`:

- Add `mode: "standard"` (framework-standard WS API with default buffer sizes, matching the Standard constraints)
- Add `echo-ws-pipeline` and `echo-ws-limited` to `tests`
- `enabled` stays `true` (kept since the last re-enable)

## Notes

- All three WS profiles share the same `/ws` endpoint; no new code is required.
- Hical's WS core `receiveInternal` reads the kernel buffer with a single `async_read_some`, drains multiple frames via a `for(;;)` loop, and consumes them with `memmove` — zero extra framework change for the pipelined echo scenario.
- The handshake stack computes `Sec-WebSocket-Accept` per connection from the client key per RFC 6455, satisfying the echo-limited constraint against pre-computed 101 responses.
- A re-benchmark run is needed for these profiles' numbers to appear on the board.


Hical61:
Quick note before review: this PR isn't final yet. Hical is adding async / async-db / latency profiles and moving type to emerging, all landing with v2.7.0. No merge for now — I'll update the PR when that's ready. Thanks.

MDA2AV:
To be considered emerging you need to first of all subscribe to all scored tests in the section so 8gbit and json-tls. 

We don't accept new applications to emerging/flagship that don't cover these, then we can evaluate whether this is experimental or emerging, for emerging you need at least some production proof using this framework and we will do a thorough review on the framework source code to understand its level of maturity and whether it is properly maintained. Once that is done I will also inform whether this can be considered standard or tuned and whether we will apply any score penalizations due to any feature lacking.

Hical61:
Thanks for the review. We don't have production usage to point to yet, so I'll move the entry to experimental for now. TLS is already supported in the framework (enableSsl + ssl::stream, with a test_ssl_connection test), so I'll add 8gbit and json-tls when applying for emerging later.

Hical61:
The entry is moved to experimental as discussed, and everything else looks good — ready to merge whenever you are. Thanks!

Hical61:
Fixed the 8gbit shortfall (TLS listener was capped at 2 threads; now sized from cgroup CPU count). Re-running to confirm.
/benchmark -f hical --save

Hical61:
8gbit fixed (0.82 → 0.99, TLS listener now sized from cgroup CPU), json-tls got the same boost. latency-500k-8cpu staying ~0.86 is expected — reference-only, plaintext single-core ceiling, not TLS-related. Ready to merge. Thanks!

PR #1484, opened 2026-09-18
Upgrade Helidon to 27.0.0

tomas-langer:
Upgrade Helidon to 27.0.0


tomas-langer:
Please do not merge yet, investigating some of the regressions

tomas-langer:
I have reverted the tuned static content change (as the tuned version is still much faster than our production impl). 
Also a small improvement using a different charset when converting numbers to bytes.

